PR #1, opened 2020-09-14
Discussion of values with units

swalker2m:
Trying to decide how to handle the fundamental problem of values with units.  Examples I find are all simplistic.  They assume, for example, floating point values in particular units.  I think we want to be more careful about units.  Here is a discussion as PR for comments.


swalker2m:
Screen Shot 2020-09-24 at 16 39 02

swalker2m:
Screen Shot 2020-09-24 at 17 24 30

PR #16, opened 2020-10-07
Strip nested data

swalker2m:
Sends events without nesting `data` fields.  For example

```
{
  "type" : "data",
  "id" : "1",
  "payload" : {
    "data" : {
      "targetEdited" : {
        "oldValue" : {
          "name" : "Betelgeuse"
        },
        "newValue" : {
          "name" : "Zero"
        }
      }
    }
  }
} 
```

instead of

```
{
  "type" : "data",
  "id" : "1",
  "payload" : {
    "data" : {
      "data" : {
        "targetEdited" : {
          "oldValue" : {
            "name" : "Betelgeuse"
          },
          "newValue" : {
            "name" : "Zero"
          }
        }
      }
    }
  }
} 
```

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/api/service/Subscriptions.scala
+++ b/modules/service/src/main/scala/lucuma/odb/api/service/Subscriptions.scala
@@ -75,8 +75,16 @@ object Subscriptions {
   // Converts raw graphQL subscription events into FromServer messages.
   private def fromServerPipe[F[_]](id: String): Pipe[F, Either[Throwable, Json], FromServer] =
     _.map {
-      case Left(err)  => Error(id, err.format)
-      case Right(obj) => Data(id, DataWrapper(obj))
+      case Left(err)           => Error(id, err.format)
+      case Right(json)         =>
+        Data(id,
+          DataWrapper(
+            // Sangria provides subscription results wrapped in "data".  If we
+            // add it directly to a DataWrapper then the encoding will have
+            // nested "data" fields.  This will strip the outer "data".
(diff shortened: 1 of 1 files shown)


modules/service/src/main/scala/lucuma/odb/api/service/Subscriptions.scala
```diff
@@ -75,8 +75,16 @@ object Subscriptions {
...
-      case Left(err)  => Error(id, err.format)
-      case Right(obj) => Data(id, DataWrapper(obj))
+      case Left(err)           => Error(id, err.format)
+      case Right(json)         =>
+        Data(id,
+          DataWrapper(
+            // Sangria provides subscription results wrapped in "data".  If we
+            // add it directly to a DataWrapper then the encoding will have
+            // nested "data" fields.  This will strip the outer "data".
+            json.mapObject(o => o("data").fold(o)(_.asObject.fold(o)(identity)))
```
swalker2m, line 85: This is a bit awkward, to say the least.  Maybe there's a better way?

rpiaggio: I don't know if it's better, but you could do `json.map(_.asObject.flatMap(_("data")))` and then either `.getOrElse(json)`, or maybe return an `Error` if `data` is not present?

I've evaluated removing `DataWrapper` from the hierarchy, but I'm not fully convinced: it would place the burden on the caller of creating the `data` level, leading to errors if omitted. Whereas this way it's in the model. We could add a `DataWrapper.fromJson` static method which would do the same as is done here, albeit it would be a bit more hidden.

rpiaggio (approved):
LGTM

PR #17, opened 2020-10-09
Added edit event filtering on id

swalker2m:
Instead of receiving every edit event, you get events for particular items.


swalker2m:
An unresolved problem is firing an initial event with the current value.  At the moment that won't necessarily happen until something actually changes.

rpiaggio (approved):
While not entirely familiar with Sangria, definitely LGTM.

PR #23, opened 2020-10-16
Pid-matching subscriptions

swalker2m:
An update to subscriptions to accept a program id argument such that only events that impact the associated program are received by the client.  For example, only events for the targets referenced by a particular program.


rpiaggio (approved):
LGTM (modulo Sangria).

PR #24, opened 2020-10-16
Pull up data stripping

swalker2m:
Sangria places results in a JSON object with a `data` field.  When we add results to a streaming message `DataWrapper` we end up with nested `data` fields.  I'd fixed this for subscription results, but we need this at a higher level to handle one-off queries over a web socket connection.


rpiaggio (approved):
LGTM

PR #26, opened 2020-10-16
target => observations

swalker2m:
As requested, adds an `observations` field to `Target`.  In this model, an observation has an optional asterism.  If the asterism is defined then it has a `Set` of targets.  So this will return all observations having an asterism that refers to this target.

@rpiaggio  Since targets (and asterisms too for that matter) can be shared across programs, do we want to add an optional program id argument and filter?

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
@@ -3,7 +3,7 @@
 
 package lucuma.odb.api.repo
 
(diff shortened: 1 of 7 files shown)


rpiaggio:
Thank you! Yes, please, we will need to get all the targets for a particular program.

swalker2m:
>> Since targets (and asterisms too for that matter) can be shared across programs, do we want to add an optional program id argument and filter?

> Yes, please, we will need to get all the targets for a particular program.

Added an optional program id filter.

Screen Shot 2020-10-19 at 10 20 20

rpiaggio (approved):
LGTM

PR #27, opened 2020-10-19
More event filters

swalker2m:
Two loosely related changes:

1. Event subscriptions now take two (optional) parameters:

   * Object id (e.g. target id for target events).  When specified only events for the particular object are sent.  This allows watching a single target, asterism, observation, or program.
   * Program id.  When specified only events for objects referenced by the associated program are sent.  This makes it possible to watch events for any target, asterism, or observation referenced by a program.

2. Renaming of id fields for consistency.  Now all ids are prefixed by the type of object in accordance with their `Gid` prefix.  For example, everywhere a program id appears it is called `pid`.  This, I think, reduces confusion and makes it possible to specify two different id types in the same argument list.

The second update is a __breaking change__.  (We'd avoid this in production but until we have an initial version I say all bets are off.)

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
@@ -160,11 +160,14 @@ object AsterismModel extends AsterismOptics {
   }
 
   final case class EditDefault(
-    id:           Id,
+    asterismId:   Id,
     existence:    Option[Existence],
     explicitBase: Option[Option[CoordinatesModel.Input]],
     targetIds:    Option[Set[TargetModel.Id]]
   ) extends Editor[Id, AsterismModel.Default] {
+
+    override def id: Id =
+      asterismId
 
     override def editor: ValidatedInput[State[AsterismModel.Default, Unit]] =
       Nested(explicitBase).traverse(_.toCoordinates).map { b =>
--- a/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
@@ -18,7 +18,7 @@ import monocle.{Lens, Optional, Prism}
 
 sealed trait AsterismModel {
 
-  def id:           AsterismModel.Id
+  def aid:          AsterismModel.Id
   def existence:    Existence
 
   def explicitBase: Option[Coordinates]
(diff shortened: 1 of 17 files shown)


rpiaggio:
For change 2:

Does Sangria require - or makes things easier if - field and parameter names to be the same? If not, may I suggest we keep the field names as `id` but have query parameters called `programId`, `targetId`, etc?

swalker2m:
> Does Sangria require - or makes things easier if - field and parameter names to be the same?

It isn't required, but there are cases where there are multiple id-type field names.  I was calling the "main" one `id` and the other one `pid` or whatever but then you always have to think whether to use `id` or `pid` so it seemed easier to just use the prefix character always.

Now I'm thinking perhaps it would be best to just always spell it out (e.g, `programId` everywhere instead of `pid`).  What do you think @rpiaggio ?

rpiaggio:
I personally prefer it to be spelled out, for the sake of readability and future-proofing against multiple entities starting with the same letter.

Also, if there are multiple `id` fields for the same entity, I think it's OK to keep calling `id` to the "primary key" for that entity, but I don't have a strong opinion here, spelling it out is fine for me too.

In summary, thinking of how queries would look like:
```
subscription {
  entityEdited(id: "e-X", programId: "p-2") {
    newValue {
      id
      programId
    }
  }
}
```

swalker2m:
@rpiaggio Thanks for your help.  I think your suggestion is better than what I had.  I'm still not completely sure.  Now the only place were the raw `id` name appears (other than event ids) is in the top-level item query.  For example:

```
query Program {
  program(programId: "p-2") {
    id
    name
    targets(includeDeleted: true) {
      id
      name
    }
  }
}
```

I could see making the parameter to the `program` query be named `id` there, but it seems better to me for it to just _always_ be spelled out.  For example, here are two subscriptions:

```
subscription TargetCreated {
  targetCreated(programId: "p-2") {
    value {
      id
      name
    }
  }
}

subscription TargetEdited {
  targetEdited(targetId: "t-2") {
    newValue {
      id
      name
    }
  }
}
```

The second would be `id` instead of `targetId` if we use `id` when referring to the primary type of item.  Then you'd have to think about which one to use in each case.  It's a minor thing but seems important to get it right.  On the other hand, we can always change things after we get some experience with it.

toddburnside:
This seems like a good approach to me. Within an object, `id` is fairly obviously the id of that object. But, in the query parameters, I agree that being explicit is clearer.

rpiaggio (approved):
Looks great to me!

PR #29, opened 2020-10-20
connection keep alive

swalker2m:
This PR introduces connection keep alive messages from the server.  There are a couple of caveats. 

## Timeout Period

Rather arbitrarily, it uses a 5 second keep alive message interval.  There seems to be no standard agreeing upon the timeout period between client and server and at any rate the underlying web socket ping-pong frames are independent of this setting.  Longer settings like 2 minutes cause the Playground to timeout and stop subscriptions.  It's unclear what value makes sense.

## Cleanup

An outstanding issue with the Playground app is that when the user stops a subscription, it dutifully sends a `stop` streaming message but never actually terminates the connection.  If the user starts another subscription it sets up another web socket connection instead of reusing the existing one.  This means that after a series of start/stop subscription actions the server is sending multiple keep alive messages back to the same client though existing connections will never be used again.

A client refresh or closing of the tab does result in stopping the keep alive messages at least.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
+++ b/modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
@@ -4,10 +4,11 @@
 package lucuma.odb.api.service
 
 import lucuma.odb.api.service.ErrorFormatter.syntax._
+import _root_.fs2.Stream
 import _root_.fs2.concurrent.Queue
 import cats.MonadError
 import cats.data.ValidatedNel
-import cats.effect.ConcurrentEffect
+import cats.effect.{ConcurrentEffect, Timer}
 import cats.implicits._
 import clue.model.StreamingMessage.FromServer.{Data, DataWrapper}
 import clue.model.StreamingMessage.{FromClient, FromServer}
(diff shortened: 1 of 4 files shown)


modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
```diff
@@ -4,10 +4,11 @@
 package lucuma.odb.api.service
 
 import lucuma.odb.api.service.ErrorFormatter.syntax._
+import _root_.fs2.Stream
```
cquiroz, line 7: Why do you need the `_root_`. Do we have an internal `fs2` package?

cquiroz (approved):
LGTM. are you doing anything with the client's response?

modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
swalker2m: You're right it isn't necessary here.  That was a habit from when I had placed the FS2/Sangria streaming integration in an `fs2` package there.

swalker2m:
> are you doing anything with the client's response?

As far as I can tell, there is no client response in the [GraphQL over WebSocket protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#session-init-phase).  As you pointed out separately, at a lower level there are websocket ping/pong messages sent back and forth but here it is purely server -> client.

PR #30, opened 2020-10-20
spell out microarcseconds

swalker2m:
A minor consistency update to use "microarcseconds" everywhere rather than a mix of "microarcseconds" and "microarcsecs".


cquiroz (approved):
LGTM

PR #36, opened 2020-10-22
Allow ids to be provided by client during creation

swalker2m:
Pending an update to supply the client with an id-pool for use (as discussed), this PR allows ids to be provided by the client.  Duplicates are not yet checked for program ids, but this is needed quickly for work on a demo today so I'm sending it as-is and will fix it up subsequently.


rpiaggio (approved):
LGTM

PR #39, opened 2020-10-23
Fix Data message creation

swalker2m:
Upgrades `clue` to pick up the `DataWrapper` with `errors` field and uses it to pass errors when present.  This also solves the nested `data` field issue in a better way.


rpiaggio (approved):
LGTM

PR #44, opened 2020-10-27
Event refactor

swalker2m:
Collapses created and updated events into one.  Simplifies logic surrounding inserts.  This is a breaking change.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
@@ -210,7 +210,7 @@ object AsterismModel extends AsterismOptics {
   final case class AsterismEvent (
     id:       Long,
     editType: Event.EditType,
-    value:    AsterismModel,
+    value:    AsterismModel
   ) extends Event.Edit[AsterismModel]
 
   object AsterismEvent {
--- a/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
@@ -207,26 +207,17 @@ object AsterismModel extends AsterismOptics {
       case (_, _)                   => false
     }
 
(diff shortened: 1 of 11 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
```diff
@@ -207,26 +207,17 @@ object AsterismModel extends AsterismOptics {
...
-  }
-
-  final case class AsterismEditedEvent (
+  final case class AsterismEvent (
     id:       Long,
-    oldValue: AsterismModel,
-    newValue: AsterismModel
-  ) extends Event.Edited[AsterismModel]
+    editType: Event.EditType,
+    value:    AsterismModel,
```
cquiroz, line 213: trailing comma

modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
```diff
@@ -12,6 +12,7 @@ import cats.data._
 import cats.effect.concurrent.Ref
 import cats.syntax.apply._
 import cats.syntax.applicative._
+import cats.syntax.either._
```
cquiroz, line 15: Any reason not to just import `cats.implicits._`?

cquiroz (approved):
I can't really comment on the details but it looks good and we discussed on the chat about removing the `oldValue` field

modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
swalker2m: Not really.  I just had a vague sense that a la carte imports are preferred.  What is the consensus?

modules/core/src/main/scala/lucuma/odb/api/model/AsterismModel.scala
swalker2m: Thanks. :+1:

PR #45, opened 2020-10-28
Target observation link

swalker2m:
I'm part-way through refactoring sharing/unsharing, but the change set is already pretty large and I have something that might be usable for the requested feature of target/observation sharing.  There will be a follow-up PR with the remainder of the refactoring which should make the various sharing options more similar.  For now, this will add `shareTargetsWithObservations` and `unshareTargetsWithObservations`.  The long-term fate of this is questionable because it requires working with asterisms, which for the moment must be `Default` asterisms.


rpiaggio (approved):
Although I don't really follow the code, I deployed a review app and played with it and it works great, thank you! LGTM

PR #47, opened 2020-10-29
Catalog ID

swalker2m:
Adds the catalog id to `Sidereal`.


modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
```diff
@@ -274,6 +279,13 @@ trait TargetOptics { self: TargetModel.type =>
   val siderealTracking: Optional[TargetModel, SiderealTracking] =
     lucumaTarget.composeOptional(gemTargetSiderealTracking)
 
+  // Add to `core` `SiderealTrackingOptics`
+  private val siderealTrackingCatalogIdLens: Lens[SiderealTracking, Option[CatalogId]] =
+    Lens[SiderealTracking, Option[CatalogId]](_.catalogId)(a => b => b.copy(catalogId = a))
```
swalker2m, line 284: I'll switch to the one I added in `core` as part of upgrading to the latest core.

modules/core/src/main/scala/lucuma/odb/api/model/syntax/OptionalEditorOps.scala
```diff
@@ -7,12 +7,18 @@ import cats.data.State
 import monocle.Optional
 import monocle.state.all._
 
-final class OptionalEditorOps[S, A](self: Optional[S, A]) {
+final class OptionalEditorOps[S, A](val self: Optional[S, A]) extends AnyVal {
```
swalker2m, line 10: Oops, I didn't mean for this to sneak in but I suppose it is harmless.  I'll be deleting in favor of similar syntax being added to `core` anyway.

rpiaggio (approved):
LGTM

PR #48, opened 2020-10-29
fire observation event

swalker2m:
As requested, adds an observation event when a target is shared with an observation.


rpiaggio (approved):
Thank you! LGTM

PR #50, opened 2020-10-29
Tracking changes to lucuma-core

swalker2m:
A mostly mechanical update that switches to use `lucuma-core` ids.  Also uses `lucuma-core` optics syntax and `catalogId` `Lens`.


cquiroz (approved):
LGTM

PR #53, opened 2020-11-02
Add target magnitude

swalker2m:
Adds target magnitude schema.  Also makes program id list optional when creating targets.


rpiaggio (approved):
LGTM

PR #58, opened 2020-11-03
Update streaming message sent for error

swalker2m:
When there is no `data`, send an `error` message instead of a `data` message with no `data` field.


rpiaggio (approved):
LGTM

PR #60, opened 2020-11-04
More Sharing

swalker2m:
A minor update to sharing (though it will break anything other than target/observation sharing).  Makes the asterism/program and target/program sharing API similar to the target/observation (plural on each side of the share) and moves them all into the `SharingMutation`.


rpiaggio (approved):
LGTM

PR #62, opened 2020-11-05
First pass at unset

swalker2m:
This is a work-in-progress since there are other editors that would have to be updated.  The idea here is to permit unsetting previously set optional values with a natural-ish syntax.  For example, if `parallax` has been set to a value you can create an input object for changing it like this:

```
{
  "editSidereal": {
    "targetId": "t-86",
    "parallax": { "microarcseconds": 1 }
  }
}
```

If you want to leave it alone and only change the target name then you just don't mention it:

```
  "editSidereal": {
    "targetId": "t-86",
    "name": "Biff Henderson"
  }
```

If you want to _unset_ the optional `parallax` value then you use a JSON `null` value as in:

```
  "editSidereal": {
    "targetId": "t-86",
    "parallax": null
  }
```

As you can see the downside is that a custom `Decoder` is necessary (I think).  What do we think of all of this?


rpiaggio:
I like the simplicity of the resulting API and am trying to think how this would look on the enconding end.

We somehow need a tri-state sum type, which could be an `Option[Option[A]]` or a new sum type, say `Set`, with possible values `Skip`, `Unset` and `Update(v)`.

However, the clue macro that generates input classes from the schema should know how to define input properties with this type instead of a simple `Option[A]`.

One way to go is to define `scalar`s in the schema for all base types (`SetString`, `SetInt`, etc.) and somehow input types should also follow this pattern.

rpiaggio:
Actually, we may need to expand the mechanism to contemplate modifying arrays (I mean schema arrays, which could be backed by List, Set, Map, etc.). We may want to replace the whole array, or to add/remove/modify elements in it.

I'm thinking of the magnitudes editor here for example, where the user can add/remove/modify entries, but we can also query the catalog in which case we want to reset the whole array.

swalker2m:
Editing should be as intuitive as possible for end users, but there is a fundamental difference between creating/replacing something vs. updating something.  Namely when creating you specify all values that you want to have set and ignore any you want to leave unset.  When updating, ideally you should only have to specify the changes you want to make.  We already have specific mutations for `createFoo` vs. `updateFoo`.  The issue though is what to do with the fields of the input objects for `updateFoo`.  All are optional fields because the idea is that you only specify the items that you want to change.  Some of the fields represent data that is itself optional like a sidereal target's parallax.  Galaxies are so far away that parallax isn't a factor.  So that begs the question, how do you update to *unset* an optional value when there are two interpretations for a missing value: (1) unset it or (2) ignore it.

Some options that occur to me:

1) Unset with an explicit empty value, either `null` or an empty object:

```
"parallax": {}
```

or

```
"parallax": null
```

2) Make the update input objects distinct from the create input objects and add a directive field like "edit" that says what to do:

```
"parallax": {
  "edit": "SET",
  "milliarcseconds": 3.2
}
```

or

```
"parallax": {
  "edit": "UNSET"
}
```

The "edit" directive could be optional and could default to an appropriate value depending on whether there is a value like `milliarcseconds` or if it is an empty object `{}`.  Clue could use the presence of a directive to generate appropriate code though?


3) Do away with optional values?  Is this always possible?  I suspect no.  At any rate, in the case of parallax targets would essentially be created with 0 parallax if not specified instead of an empty parallax.  If you want to "unset" it you have to explicitly specify a parallax of 0.

```
"parallax": {
  "microarcseconds": 0
}
```

4) In the case of update input objects, add explicit `unset` fields:

```
"unsetParallax": true
```

Presumably specifying both a value and `unset` would trigger an error:

```
"parallax": { "microarcseconds": 100 }
"unsetParallax": true
```

I think option (1) is easy to understand but there's no intuitive way in the schema to distinguish fields like "parallax" which are optional from "name" which are not.  This wouldn't work for example:

```
"name": null
```

Any other ideas?

swalker2m:
Fields like `magnitudes` which are arrays of values present a challenge as well as Raul points out.  Options are similar to the object fields like `parallax`.

(1) Distinguish the create and update `input` objects and add a directive:

```
"magnitudes": {
  "edit": "SET",
  "values": [
    {
      "band": "K",
      "value": 13.3,
      "system": "VEGA"
    }
  ]
}
```

where `edit` can be `SET` or `UPDATE`.

(2) Add an `unsetMagnitudes` boolean field and make it an error to specify both `magnitudes` and `unsetMagnitudes`.

(3) Punt, in this case at least, and make a `Magnitudes` object with fields named for each magnitude bad.  In other words, make this just an object like parallax is an object:

```
"magnitudes": {
  "K": {
    "value": 13.3
    "system": "VEGA"
  }
}
```

Then we adopt whatever we do for `parallax`.  This doesn't apply for all arrays, but I think it works pretty well for this case.  Seems intuitive for a user.

Other ideas?

swalker2m:
Updated to

1) Support empty objects for removal as well as `null`, eg: `"parallax" : {}`
2) Completely remove `properVelocity` instead of just deprecating it.
3) Add an option for magnitude editing with fields:
   * `magnitudes` - to (re)set all magnitude values
   * `modifyMagnitudes` - to add/update existing magnitude list
   * `deleteMagnitudes` - to remove particular magnitude bands

swalker2m:
While not exactly describing the same situation, [this section from the spec](https://spec.graphql.org/June2018/#sec-Null-Value) seems to support the idea of using `null` to represent a request to delete a value.

> For example, these two field calls are similar, but are not identical:
>
> Example № 29
```
{
  field(arg: null)
  field
}
```
>
>The first has explictly provided null to the argument “arg”, while the second has implicitly not provided a value to the argument “arg”. These two forms may be interpreted differently. For example, a mutation representing deleting a field vs not altering a field, respectively. Neither form may be used for an input expecting a Non‐Null type.

rpiaggio (approved):
LGTM, we are using it successfully from the UI

PR #63, opened 2020-11-09
debugging dots of death

swalker2m:
Adds a bit of debugging info in an attempt to collect more information about an issue where mutations hang in the API service, causing "spinning dots of death" in the UI.


cquiroz (approved):
LGTM

PR #65, opened 2020-11-09
Library updates

swalker2m:
Library updates.  Rename (and deprecate) `properVelocity`, favoring `properMotion` instead.


build.sbt
```diff
@@ -5,13 +5,13 @@ val catsVersion                 = "2.2.0"
...
 val clueVersion                 = "0.3.4"
-val fs2Version                  = "2.4.4"
+val fs2Version                  = "2.4.5"
 val http4sVersion               = "0.21.8"
 val jawnVersion                 = "1.0.0"
 val kindProjectorVersion        = "0.11.0"
 val logbackVersion              = "1.2.3"
 val log4catsVersion             = "1.1.1"
-val lucumaCoreVersion           = "0.6.5"
+val lucumaCoreVersion           = "0.6.6"
```
cquiroz, line 14: There is 0.7.0 already

modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
```diff
@@ -141,7 +141,8 @@ trait TargetMutation extends TargetScalars {
   val InputObjectTypeCreateSidereal: InputObjectType[TargetModel.CreateSidereal] =
     deriveInputObjectType[TargetModel.CreateSidereal](
       InputObjectTypeName("CreateSiderealInput"),
-      InputObjectTypeDescription("Sidereal target parameters")
+      InputObjectTypeDescription("Sidereal target parameters"),
+      DocumentInputField("properVelocity", "Deprecated, use properMotion instead.")
```
swalker2m, line 145: This is not very satisfactory, but there seems to be no way to mark an input field deprecated.

swalker2m:
In theory, this is a non-breaking change.  Once Explore is updated though we should remove `properVelocity`.

cquiroz (approved):
LGTM

PR #66, opened 2020-11-10
stop spurious events

swalker2m:
When using Explore, switching from one target to another generates spurious edits.  For example, if the focus is in the target name when clicking on a different target the focus is lost.  That triggers a `mutation` request from the client whether or not the target name has actually changed.  

```
2020-11-10T13:58:34.800652+00:00 app[web.1]: [ioapp-compute-2] INFO  l.o.a.s.Routes - Received message from client: Text('{
2020-11-10T13:58:34.800668+00:00 app[web.1]: "type" : "start",
2020-11-10T13:58:34.800668+00:00 app[web.1]: "id" : "6761a6a7-2b5b-4855-9b11-c2e93cc9092e",
2020-11-10T13:58:34.800669+00:00 app[web.1]: "payload" : {
2020-11-10T13:58:34.800680+00:00 app[web.1]: "query" : "\n      mutation($input: EditSiderealInput!) {\n        updateSiderealTarget(input: $input) {\n          id\n        }\n      }\n    ",
2020-11-10T13:58:34.800681+00:00 app[web.1]: "variables" : {
2020-11-10T13:58:34.800681+00:00 app[web.1]: "input" : {
2020-11-10T13:58:34.800682+00:00 app[web.1]: "targetId" : "t-15099551",
2020-11-10T13:58:34.800682+00:00 app[web.1]: "name" : "NGC 2070"
2020-11-10T13:58:34.800682+00:00 app[web.1]: }
2020-11-10T13:58:34.800683+00:00 app[web.1]: }
2020-11-10T13:58:34.800683+00:00 app[web.1]: }
2020-11-10T13:58:34.800683+00:00 app[web.1]: }', last: true)
```

The edit in turn causes subscriptions to be notified of an update despite the fact that nothing has really changed.

```
2020-11-10T13:58:34.811797+00:00 app[web.1]: [ioapp-compute-0] INFO  l.o.a.s.Main - subscription event (1 still queued): TargetEvent(108,Updated,TargetModel(t-15099551,Present,Target(NGC 2070,Right(SiderealTracking(Some(CatalogId(Simbad,NGC  2070)),Coordinates(05:38:42.000000 -69:05:59.999999),Epoch(J2000.000),None,Some(RadialVelocity(251.0 km/s)),None)),TreeMap())))
2020-11-10T13:58:34.812441+00:00 app[web.1]: [ioapp-compute-0] INFO  l.o.a.s.Main - subscription event (1 still queued): TargetEvent(108,Updated,TargetModel(t-15099551,Present,Target(NGC 2070,Right(SiderealTracking(Some(CatalogId(Simbad,NGC  2070)),Coordinates(05:38:42.000000 -69:05:59.999999),Epoch(J2000.000),None,Some(RadialVelocity(251.0 km/s)),None)),TreeMap())))
```
... many more of the same as all clients are notified ...

This PR will avoid triggering events unless an item is actually edited according to an `Eq` comparison of the old and new values.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -7,6 +7,7 @@ import lucuma.odb.api.model.Existence._
 import lucuma.core.`enum`.ObsStatus
 import lucuma.core.optics.syntax.lens._
 import lucuma.core.model.{Asterism, Observation, Program}
+import cats.Eq
 import cats.data.State
 import cats.syntax.validated._
 import io.circe.Decoder
@@ -28,6 +29,9 @@ object ObservationModel extends ObservationOptics {
 
   implicit val TopLevelObservation: TopLevelModel[Observation.Id, ObservationModel] =
     TopLevelModel.instance(_.id, ObservationModel.existence)
+
+  implicit val EqObservation: Eq[ObservationModel] =
+    Eq.by(o => (o.id, o.existence, o.programId, o.name, o.status, o.asterismId, o.plannedTimeSummary))
 
   final case class Create(
     observationId: Option[Observation.Id],
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
@@ -7,17 +7,10 @@ import lucuma.odb.api.model.{Editor, Event, Existence, InputError, TopLevelModel
 import lucuma.odb.api.model.syntax.toplevel._
 import lucuma.odb.api.model.syntax.validatedinput._
 import lucuma.core.util.Gid
-import cats.{FunctorFilter, Monad, MonadError}
+import cats.{Eq, FunctorFilter, Monad, MonadError}
 import cats.data._
 import cats.effect.concurrent.Ref
-import cats.syntax.apply._
-import cats.syntax.applicative._
-import cats.syntax.either._
-import cats.syntax.flatMap._
-import cats.syntax.functor._
-import cats.syntax.functorFilter._
-import cats.syntax.option._
-import cats.syntax.validated._
+import cats.syntax.all._
 import cats.kernel.BoundedEnumerable
 import monocle.Lens
 import monocle.function.At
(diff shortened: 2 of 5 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -29,6 +30,9 @@ object ObservationModel extends ObservationOptics {
   implicit val TopLevelObservation: TopLevelModel[Observation.Id, ObservationModel] =
     TopLevelModel.instance(_.id, ObservationModel.existence)
 
+  implicit val EqObservation: Eq[ObservationModel] =
```
cquiroz, line 33: I suppose we could have tests of these `Eq` instances

modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
```diff
@@ -171,24 +170,27 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, ?]](
...
 
         (tables, result)
       }.flatMap(_.liftTo[F])
 
     for {
-      u <- doUpdate
-      _ <- eventService.publish(edited(Event.EditType.Updated, u))
-    } yield u
+      e <- doUpdate
+      _ <- e.fold(_ => M.unit, u => eventService.publish(edited(Event.EditType.Updated, u)))
```
cquiroz, line 192: I think this can be done with something like `eventService....whenM(e.isDefined)`. I don't know if it really helps

cquiroz (approved):
LGTM, minor comments inside

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: Yeah you're right. It's temporary code but it may be around for a while and I've avoided it long enough I think.  I'll go back and add some arbitraries, etc.

modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
swalker2m: What would this look like exactly?  `e` is an `Either[U, U]` btw, basically using left vs. right to record unchanged vs. changed along with the correct value of the `U` object.

cquiroz: Oh, I thought it was a `Monad`, please disregard

PR #70, opened 2020-11-13
Example script

swalker2m:
Updates `Init` to make it easier to specify the initial targets, adds an example python script that adds additional targets.


swalker2m:
Someone who actually knows python could improve this, presumably.

cquiroz (approved):
Looks great, thanks

PR #73, opened 2020-11-18
Target Eqs, Arbs, and Cogs

swalker2m:
A spinach-eating, soul-crushing collection of `Eq`, `Arbitrary` and `Cogen` definitions for `TargetModel` and its friends.  Not really used to do much testing (beyond `EqTests`) yet but it does enable future testing.  A valid complaint is that all the arbitraries produce valid inputs (i.e., nothing yet that should produce an error if received as an actual input).


modules/core/src/main/scala/lucuma/odb/api/model/ParallaxModel.scala
```diff
@@ -3,9 +3,10 @@
 
 package lucuma.odb.api.model
 
+import cats.Eq
 import lucuma.core.math.{Angle, Parallax}
 import lucuma.core.util.{Display, Enumerated}
-import cats.syntax.validated._
+import cats.implicits._
```
rpiaggio, line 9: I was under the impression we were moving to `import cats.syntax.all._`.

rpiaggio (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/model/ParallaxModel.scala
swalker2m: Thanks for spotting that.  It was a wrong turn so I've put it back the way it was.

PR #79, opened 2020-11-24
Use lucuma-sso to determine the User

swalker2m:
This PR adds code to use SSO in validating and extracting user information from requests.  It requires setting up a bit of configuration (now using [ciris](https://cir.is)):

* `ODB_SSO_ROOT` - URL of the SSO service
* `ODB_SSO_PUBLIC_KEY` - Public key of the SSO service
* `ODB_SERVICE_JWT` - The JWT used by the service itself for communication with other services

At the moment, nothing is actually __done__ with the `User` and all requests proceed as before.  The only change is that the user associated with a request is logged.

For the most part, incorporating SSO is straightforward following the backend client example in `lucuma-sso`.  The difficulty comes with websocket GraphQL subscriptions, which delay providing authorization information until the `connection_init` message.  That prompted some refactoring and the bulk of the updates.


cquiroz (approved):
LGTM

PR #96, opened 2020-12-07
Library updates + eq fix

swalker2m:
* library updates
* a minor adjustment to track `clue` `0.5.0`
* missing fields for `Eq` implementation


modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
```diff
@@ -298,7 +300,7 @@ object Connection {
 
         override def receive(m: FromClient): F[Unit] =
           m match {
-            case ConnectionInit(m)   => init(m)
+            case ConnectionInit(m)   => init(m.toMap)
```
rpiaggio, line 303: In clue `0.6.0`, this will be a `Map` instead of a `JsonObject`.

rpiaggio (approved):
LGTM. You may want to wait for clue `0.6.0` so as not to have to change again.

PR #101, opened 2020-12-10
Less logging

swalker2m:
A few minor logging updates to avoid flooding the log:

1) Lowers the priority of keep alive messages to debug.
2) Uses `log4cats` everywhere.
3) Redacts authorization information coming in websocket `connection_init` message `payload`.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
+++ b/modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
@@ -6,7 +6,6 @@ package lucuma.odb.api.service
 import lucuma.odb.api.service.ErrorFormatter.syntax._
 import lucuma.core.model.User
 import lucuma.sso.client.SsoClient
-import lucuma.odb.syntax.logger._
 import cats.MonadError
 import cats.data.ValidatedNel
 import cats.effect.{ConcurrentEffect, Timer}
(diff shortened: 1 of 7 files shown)


cquiroz (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
```diff
@@ -100,12 +103,45 @@ object Routes {
...
+              info(u, s"Sending to client: ${trimmedMessage(msg)}")
+          }
+        }
+
+      def logTextFrame(user: F[Option[User]], f: WebSocketFrame): F[Unit] = {
+
+        // The connection_init message payload has authorization information
+        // which should not be logged.
+        val AuthRegEx    = """"Authorization":\s*"[^"]*"""".r.unanchored
+        val RedactedAuth = """"Authorization": <REDACTED>"""
```
rpiaggio, line 128: This is a super tiny detail, but if you don't want to repeat `Authorization` I think you can capture it in `AuthRegEx` (`("Authorization":)`) and then use `$1` in `RedactedAuth`.

rpiaggio (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/api/service/Routes.scala
swalker2m: Thanks, that was a good idea.  :+1:

PR #115, opened 2020-12-23
Unset asterism, etc. with null value

swalker2m:
Adds the ability to disconnect observations and asterisms (via `updateObservation`, passing in `null` for `asterismId`).  Switches to clue `Input` in general for observation editing.


cquiroz (approved):
LGTM

PR #131, opened 2021-01-28
Asterism Updates

swalker2m:
@rpiaggio @cquiroz  This is an unreviewable PR that implements breaking API changes we discussed:

* Observation now references an `Option[Either[Asterism.Id, Target.Id]]` which is revealed in the GraphQL schema as an `asterism: Option[Asterism]` and a `target: Option[Target]` though at most one will ever have a value.
* Observation creation and edit updated accordingly to allow directly setting either an asterism or a target.
* Asterism subtypes (`Default` and `Ghost`) have been eliminated in favor of the idea that an asterism is a collection of targets with (eventually) optional instrument specific configuration bits.
* Sharing mutations updated accordingly.

I won't merge this PR without prior approval since it will break the staging Explore.


rpiaggio:
Thank you @swalker2m . I will deploy a review app so that we can start migrating to this schema and when ready we can sync merge.

Can we make the schema publish a union type `union Targets = Asterism | Target` ?

rpiaggio:
Just for everybody's reference: https://lucuma-odb-p-obs-target-1hnxxk.herokuapp.com/playground.html

swalker2m:
> Can we make the schema publish a union type `union Targets = Asterism | Target` ?

That seems like a good idea.  I'm not sure about the name.  [Here it is](https://github.com/gemini-hlsw/lucuma-tmp-api/pull/131/commits/11aaaf2ec9013ae960fee92aa1276826105743e0) as `ObservationTarget` :-/

PR #148, opened 2021-02-10
Add ConstraintSetModel

toddburnside:


modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
```diff
@@ -0,0 +1,189 @@
...
+import lucuma.core.optics.syntax.lens._
+import lucuma.odb.api.model.Existence._
+import lucuma.odb.api.model.syntax.input._
+import monocle.{ Fold, Lens, Optional }
+import monocle.macros.GenLens
+
+final case class ConstraintSetModel(
+  id:              ConstraintSet.Id,
+  existence:       Existence,
+  programId:       Program.Id,
```
swalker2m, line 24: Having a program id here means that unlike targets, we don't anticipate sharing constraint sets across programs.  I agree that's a fine assumption to start with.

modules/core/src/main/scala/lucuma/odb/api/model/ValidatedInput.scala
```diff
@@ -15,19 +18,33 @@ object ValidatedInput {
...
     }
 
   def nonEmptyString(name: String, s: String): ValidatedInput[NonEmptyString] =
     NonEmptyString
       .from(s)
       .leftMap(err => InputError.fromMessage(s"'$name' may not be empty: $err"))
       .toValidatedNec
+
+  def closedInterval[T](name: String, value: T, low: T, high: T)(implicit
+    v:                        Validate[T, Interval.Closed[low.type, high.type]]
```
swalker2m, line 45: Scalafmt?

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
```diff
@@ -0,0 +1,189 @@
...
+
+  final case class Create(
+    constraintSetId: Option[ConstraintSet.Id],
+    programId:       Program.Id,
+    name:            String,
+    imageQuality:    ImageQuality,
+    cloudExtinction: CloudExtinction,
+    skyBackground:   SkyBackground,
+    waterVapor:      WaterVapor,
+    elevationRange:  ElevationRangeModel.Create
```
swalker2m, line 48: Should a user be required to specify all of these fields in order to create a constraints set?  Should any of them be optional?  It seems like we could default most of the values?

modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala
```diff
@@ -0,0 +1,219 @@
...
+  implicit val EqAirmassRange: Eq[AirmassRange] = Eq.fromUniversalEquals
+
+  /**
+   * Input parameter used to create an airmass range. Parameter ranges
+   * are validated and deciMin must be <= deciMax
+   */
+  final case class Create(deciMin: Int, deciMax: Int) {
+    def create: ValidatedInput[AirmassRange] =
+      (ValidatedInput.closedInterval("deciMin", deciMin, 10, 30),
+       ValidatedInput.closedInterval("deciMax", deciMax, 10, 30)
```
swalker2m, line 97: A goal of the API is to make it possible to enter values in units that would seem most natural to the user.  See the `WavelengthModel` or `RightAscensionModel` for example.  In this case, my understanding is that astronomers will think of airmass in decimal values where `1.0` is the minimum (zenith).  If I understand it, we'd be asking them to convert this to 10 X the airmass for data entry?

swalker2m (commented):
I have some comments for discussion but I think it looks great! It's is excellent to have someone to help out with this stuff.

modules/core/src/main/scala/lucuma/odb/api/model/ValidatedInput.scala
toddburnside: Sigh. Yeah. I think the scalafmt config we use for Explore is too aggressive in aligning things and I just copied it to use here. I'll try some things to make it less awful for this case. Speaking of which, if we aren't using a common scalafmt in this project, should I add `.scalafmt.conf` to the `.gitignore` file?

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
toddburnside: I went back and forth on that. In the UI, they will essentially be forced to enter them because they will be defaulted in the dropdowns. Well, except for the elevation range which will be decimal inputs, but those would also have "defaults". But, for users of the API, I guess maybe we should provide defaults for them. I'll ask Andy what those should be.

modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala
toddburnside: I was trying to keep exact values for the storage values. But, I guess BigDecimal would actually provide that for this situation. Entry in the UI would have been in decimals, and I was going to provide alternative units for the API. But, changing to BigDecimal, or even Double, would probably be sufficient for this case. (I just have a knee jerk reaction to floating point numbers.)

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
toddburnside: Yeah. I guess it was an assumption that they wouldn't be shared between programs. I can ask Andy, though.

modules/core/src/main/scala/lucuma/odb/api/model/ValidatedInput.scala
swalker2m: Yeah, that's probably a good idea.  I'm not convinced it makes the formatting better.

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
swalker2m: Right, for example `waterVapor` won't matter as I understand it for optical observations.  Ideally you would only have to specify values you care about.

modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala
swalker2m: Agree about the reaction to floating point and about storing them as you have it.  I think it's fine though to allow API users to send decimal values.  Though it is more work, what I've been doing is supporting all natural units.  The smallest possible useful resolution of something is specified as an integer/long value and others as `BigDecimal`.

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
toddburnside: Andy has stated that constraint sets won't be shared across programs.

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
toddburnside: Andy says it's fine to require all values in the API. I told him we can switch to allowing defaults later if he changes his mind.

modules/core/src/main/scala/lucuma/odb/api/model/ValidatedInput.scala
toddburnside: I have tweaked the scalafmt config file I'm using to format the above better. I will push a commit with that change. I will use scalafmt on files I create, but try to manually format changes to existing files.

I don't think anyone would argue that automated formatting is better than artisanally crafted formatting. The advantages in my mind are:

1. Consistency in a project with multiple authors - esp. as team size gets bigger.
2. I don't need to spend time or brain cycles on formatting. I just type and let scalafmt handle it. No need to go back and fix vertical alignment if I change something, etc.

Oh, and 3), formatting is no longer a code review issue. 😆 

You can probably never get exactly what you want with scalafmt, but tweaking the file can get close. For the most egregious cases, I just do `// format: off`

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
swalker2m: We can certainly start with this.  I try to keep in mind the poor schlub sitting in front of a Playground UI typing in the stuff needed to make an observation.  There are a lot of parameters here that have to be specified even if most of them would be defaultable by making these `Option`s and then getting the least restrictive value when not specified.

modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala
swalker2m: When it is translated into a GraphQL input, it will have required `deciMin` and `deciMax` parameters by default.  That'd be fine for Explore or other UIs but doesn't seem like it would be intuitive for most human users.  I hate to suggest it but I think we might want an `AirmassModel` along the lines of the `WavelengthModel` (with integer `deciAirmass` and decimal `airmass` options?) or else to just switch to `min` and `max` `BigDecimal` airmasses.  Same for hour angle below.

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
swalker2m: Sounds good.

swalker2m (approved):
I'm fine with merging this now but before we move on too far I do think we need to add natural airmass and hour angle input options.

modules/core/src/main/scala/lucuma/odb/api/model/ValidatedInput.scala
swalker2m: I concede point 1 but I think I prefer inconsistency over what you get with scalafmt.  Points 2 and 3 are often claimed but seem to be the opposite of reality -- I suppose because of people like me who whine about what scalafmt produces.  :-)

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
toddburnside: Agreed. I can run it by Andy again before the epic is over. Maybe part of it is wanting to avoid confusion over expectations. In the UI the defaults won't be the least restrictive values.

modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala
toddburnside: I'm fine with creating something like the `WavelengthModel` for these. However, the degree of precision of BigDecimal might be OK in this case. The UI will only be allowing 1 decimal place. This doesn't seem like the kind of value we'll be doing a lot of repeated conversions with. Will we just be sending it to the Constraint Solver?

swalker2m: Right, unlike wavelengths or angular separation there aren't multiple unit options that really make sense.   Making the values `BigDecimal` seems fine to me.

PR #159, opened 2021-02-17
Back out DebuggingRef

swalker2m:
Stops using `DebuggingRef`, a custom `Ref` that does some logging and handles modify retries itself.  This was part of a debugging effort for the "spinning dots of death" issue that plagued Explore in the early days.  If the spinning dots don't return then I'll remove `DebuggingRef` altogether.


cquiroz (approved):
LGTM

PR #168, opened 2021-02-20
Create ConstraintSetRepo and associated changes/tests.

toddburnside:
This PR is built on top of #148.

Naming is hard. Feel free to suggest different names for some of these things. I won't be offended.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
@@ -294,7 +294,7 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, *]: Eq]
   ): F[T] = 
     share[Id, J, M](name, id, oneId, findM, editedM){ vtm =>
       vtm.traverse { _ => 
-        eitherUpdate[Id, OneToManyUnique, Tables, J, I, InputError]((oneId, id), linkLens)(_ + _)}
+        eitherUpdate[Id, J, I]((oneId, id), linkLens)(_ + _)}
           .map(_.andThen(identity))
     }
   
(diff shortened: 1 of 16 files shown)


modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
```diff
@@ -251,8 +253,97 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, *]: Eq]
...
+    linkLens: Lens[Tables, OneToManyUnique[I, J]],
+    editedM: M => Long => Event.Edit[M]
+  ): F[T] = 
+    share(name, input.one, input.many, findM, editedM) { vtm =>
+      vtm.traverse_ { _ => linkLens.mod_(links => links -- input.tupleLeft) }.map(_.validNec[InputError].void)
+    }
+
+  private def eitherUpdate[G[_] : Traverse, H[_, _], S, A, B, E](
+    ab: G[(A, B)], lens: Lens[S, H[A, B]]
+  )(
```
toddburnside, line 337: I tried to make this function fairly generic, but that turned out to be quite difficult to do. I guess that's why I couldn't find anything "built in" to monocle.

Since it is confined to this file, I could simplify the type parameters by fixing `S -> Table` and `E -> InputError`. With it's current usage, I could even fix `A -> J`, `B -> I` and `H -> OneToManyUnique`. Unless you think it might be useful for anything else.

modules/core/src/test/scala/lucuma/odb/api/repo/OdbRepoTest.scala
```diff
@@ -73,6 +74,192 @@ trait OdbRepoTest {
       (shared == initial ++ some) && (unshared == initial -- some)
     }
 
+  def shareWithOneUniqueTest[I, A, J, B](
```
toddburnside, line 77: None of the tests in this file, including the original `sharingTest` fully test the error cases for the case where an attempt is made to share `Id`s for non-existent entities. I have some ideas for testing this, but wasn't sure if I should spend more time on it if much of this will be thrown out in the future. If you think we might keep the `*Repo` classes but give them different persistence, it might be worth adding those tests.

swalker2m: I don't expect they will be used in actual production.  The goal I had in mind was just to make the backend work well enough to support Explore development while figuring out a viable GraphQL API.

swalker2m (approved):
I took a quick look through the code and it all seemed good to me.  :+1:

PR #172, opened 2021-02-24
Paging

swalker2m:
Updates queries that return multiple results to use the [paging connection model](https://graphql.org/learn/pagination/#complete-connection-model). 

```
query ProgramIds {
  programs(first: 10, after:"cC0y") {
    edges {
      node {
        id
      }
      cursor
    }
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```


toddburnside:
> [_This is a draft since I want to add tests, but thought I should give a heads up._]

Thanks for the heads up. I'll have to incorporate this into my current work.

modules/core/src/main/scala/lucuma/odb/api/repo/ProgramRepo.scala
```diff
@@ -16,9 +15,49 @@ import cats.effect.concurrent.Ref
...
-  def selectAllForAsterism(aid: Asterism.Id, includeDeleted: Boolean = false): F[List[ProgramModel]]
-
-  def selectAllForTarget(tid: Target.Id, includeDeleted: Boolean = false): F[List[ProgramModel]]
+  /**
+   * Selects a page of results corresponding to the given asterism id.
+   *
+   * @param aid the asterism associated with all the resulting programs in the
+   *            page
+   * @param includeObservations whether to include programs only linked through
+   *                            observations that use the asterism
```
swalker2m, line 24: This feels a bit low rent but was needed to make the test cases work.  I'm not sure I want to expose it in the actual schema as a parameter though.  Basically "all programs associated with an asterism" now includes those that are not directly linked to the asterism but that are linked only via an observation that they contain (i.e., those programs that have an observation that use the asterism are included as well).  This flag will ignore asterisms only linked via an observation.

toddburnside (approved):
LGTM

PR #178, opened 2021-02-26
Handle graph QL protocol-level reinitialization

swalker2m:
This PR purports to handle reinitialization without requiring a new web socket connection.  Fair warning, it will require some testing and may not in fact, um, work.

### DIFF
--- a/modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
+++ b/modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
@@ -43,6 +43,11 @@ sealed trait Connection[F[_]] {
    */
   def receive(m: FromClient): F[Unit]
 
+  /**
+   * Close the connection, never to be heard from again.
+   */
+  def close: F[Unit]
+
 }
 
 object Connection {
@@ -72,7 +77,7 @@ object Connection {
      *
      * @return state transition and action to execute
      */
-    def init(
+    def reset(
       user: Option[User],
       send: Option[FromServer] => F[Unit],
       subs: Subscriptions[F]
@@ -93,11 +98,20 @@ object Connection {
     def stop(id: String): (ConnectionState[F], F[Unit])
 
     /**
-     * Terminates the connection, closing all ongoing subscriptions as well.
+     * Stops all the subscriptions.
      *
      * @return state transition and action to execute
      */
-    def terminate: (ConnectionState[F], F[Unit])
+    def stopAll: (ConnectionState[F], F[Unit])
+
+    /**
+     * Closes the connection, signaling that nothing more will be sent via
+     * the reply queue.
+     *
+     * @return state transition and action
+     */
+    def close: (ConnectionState[F], F[Unit])
+
   }
 
   /**
@@ -113,7 +127,7 @@ object Connection {
     override def user: Option[User] =
       None
 
-    override def init(
+    override def reset(
       user: Option[User],
       send: Option[FromServer] => F[Unit],
       subs: Subscriptions[F]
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
```diff
@@ -93,11 +98,20 @@ object Connection {
     def stop(id: String): (ConnectionState[F], F[Unit])
 
     /**
-     * Terminates the connection, closing all ongoing subscriptions as well.
+     * Stops all the subscriptions.
      *
      * @return state transition and action to execute
      */
-    def terminate: (ConnectionState[F], F[Unit])
+    def stopAll: (ConnectionState[F], F[Unit])
```
swalker2m, line 105: What used to terminate the connection now just stops all the active subscriptions.

modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
```diff
@@ -93,11 +98,20 @@ object Connection {
...
+
+    /**
+     * Closes the connection, signaling that nothing more will be sent via
+     * the reply queue.
+     *
+     * @return state transition and action
+     */
+    def close: (ConnectionState[F], F[Unit])
```
swalker2m, line 113: There's a close that is called when an actual `Close` frame is received.

modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
```diff
@@ -151,17 +169,16 @@ object Connection {
...
-        for {
-          _ <- info(u, "received connection_init on already initialized connection")
-          _ <- action
-        } yield ()
-      }
+      (new Connected(odbService, u, r, s),
+        subscriptions.removeAll  *>
+          r(Some(ConnectionAck)) *>
+          r(Some(ConnectionKeepAlive))
+      )
```
swalker2m, line 181: This is the big change.  A reset on an active connection just removes existing subscriptions and keeps on going.  There's no need to send a `TerminateConnection`, though you can if you want.

modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
```diff
@@ -299,8 +319,11 @@ object Connection {
             case ConnectionInit(m)   => init(m)
             case Start(id, request)  => handle(_.start(id, request))
             case Stop(id)            => handle(_.stop(id))
-            case ConnectionTerminate => handle(_.terminate)
+            case ConnectionTerminate => handle(_.stopAll)
```
swalker2m, line 322: `ConnectionTerminate` just stops subscriptions now.

modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
rpiaggio: If I understand correctly, `connection_init` now just resets subscriptions (and replies `ack` + `ka`). What happens if the client sends a `start` before calling `connection_init`? (or between `connection_terminate` and `connection_init`)

rpiaggio (approved):
LGTM

modules/service/src/main/scala/lucuma/odb/api/service/Connection.scala
swalker2m: If you call `start` before the first `connection_init` it will terminate the connection.  From then on a `connection_terminate` will just remove all subscriptions.  So a terminate followed by a `start` will leave you with one subscription.  Follow that will `connection_init` and it will remove the one subscription and update the user.

So really, there's no need to call `connection_terminate` unless you want to remove all subscriptions.

PR #184, opened 2021-03-04
Creation Payload

swalker2m:
Introduces a `Payload` type for the object creation mutations.  The idea is to make it easier to evolve the API by nesting the result of a mutation, in this case creation of objects, in a containing object.  The `Payload` types are unions of a successful result or an `InputError`.  For example:

```
mutation CreateObservation($createObservation: CreateObservationInput!) {
  createObservation(input: $createObservation) {
    ... on CreateObservationSuccess {
      observation {
        name
        id
      }
    }
    ... on InputErrors {
      messages
    }
  }
}
```

As you can see this raises some errors to the level of the schema itself.  In particular problems with the input that validate according to the schema but are nonetheless illegal.  For example, references to objects that don't exist or, say, an observation associated with _both_ an asterism and a target.

I'm convinced by the `Payload` concept but on the fence about the error stuff.  Comments very welcome.  If merged, the remaining mutations would get their own `Payload` result types.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
@@ -3,7 +3,7 @@
 
 package lucuma.odb.api.repo
 
-import lucuma.odb.api.model.{AsterismModel, ObservationModel, ProgramModel, Sharing, TargetModel}
+import lucuma.odb.api.model.{AsterismModel, ObservationModel, ProgramModel, Sharing, TargetModel, ValidatedInput}
 import lucuma.odb.api.model.AsterismModel.{AsterismEvent, Create}
 import lucuma.odb.api.model.syntax.validatedinput._
 import lucuma.core.model.{Asterism, Observation, Program, Target}
@@ -34,7 +34,9 @@ sealed trait AsterismRepo[F[_]] extends TopLevelRepo[F, Asterism.Id, AsterismMod
 
   def selectForObservation(oid: Observation.Id, includeDeleted: Boolean = false): F[Option[AsterismModel]]
 
-  def insert(input: AsterismModel.Create): F[AsterismModel]
+  def insert(input: AsterismModel.Create): F[ValidatedInput[AsterismModel]]
+
+  def unsafeInsert(input: AsterismModel.Create): F[AsterismModel]
 
   def shareWithObservations(input: Sharing[Asterism.Id, Observation.Id]): F[AsterismModel]
 
@@ -121,7 +123,7 @@ object AsterismRepo {
           _   <- Tables.programAsterism.mod_(_ ++ programs.toList.tupleRight(a.id))
         } yield a
 
-      override def insert(input: Create): F[AsterismModel] =
+      override def insert(input: Create): F[ValidatedInput[AsterismModel]] =
         constructAndPublish { t =>
           val existing = tryNotFindAsterism(t, input.asterismId)
           val programs = input.programIds.traverse(tryFindProgram(t, _))
@@ -130,6 +132,9 @@ object AsterismRepo {
             addAsterism(input.asterismId, input.programIds.toSet, f)
           )
         }
+
+      override def unsafeInsert(input: Create): F[AsterismModel] =
+        insert(input) >>= (_.liftTo[F])
 
       def obsSharing(
         input:   Sharing[Asterism.Id, Observation.Id],
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
@@ -3,13 +3,15 @@
 
 package lucuma.odb.api.repo
 
-import lucuma.odb.api.model.{ObservationModel, PlannedTimeSummaryModel}
+import lucuma.odb.api.model.{ConstraintSetModel, ObservationModel, PlannedTimeSummaryModel, ValidatedInput}
 import lucuma.core.model.{Asterism, ConstraintSet, Observation, Program, Target}
 import lucuma.odb.api.model.ObservationModel.ObservationEvent
+import lucuma.odb.api.model.syntax.validatedinput._
+
+import cats.Monad
 import cats.effect.Sync
 import cats.effect.concurrent.Ref
 import cats.implicits._
-import lucuma.odb.api.model.ConstraintSetModel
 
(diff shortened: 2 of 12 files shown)


modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
```diff
@@ -34,7 +34,9 @@ sealed trait AsterismRepo[F[_]] extends TopLevelRepo[F, Asterism.Id, AsterismMod
 
   def selectForObservation(oid: Observation.Id, includeDeleted: Boolean = false): F[Option[AsterismModel]]
 
-  def insert(input: AsterismModel.Create): F[AsterismModel]
+  def insert(input: AsterismModel.Create): F[ValidatedInput[AsterismModel]]
+
+  def unsafeInsert(input: AsterismModel.Create): F[AsterismModel]
```
swalker2m, line 39: The `unsafe` alternatives are helpful for setting up a program programmatically (as is the case in the `Init` code).  In this case, we just want the whole thing to fail if we've made a mistake in wiring it up.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -113,13 +117,15 @@ object ObservationRepo {
...
-        def construct(s: PlannedTimeSummaryModel): F[ObservationModel] =
+        def construct(s: PlannedTimeSummaryModel): F[ValidatedInput[ObservationModel]] =
           constructAndPublish { t =>
-            (tryNotFindObservation(t, newObs.observationId) *>
-             tryFindProgram(t, newObs.programId)            *>
-              newObs.withId(s)
+            (tryNotFindObservation(t, newObs.observationId)    *>
+             tryFindProgram(t, newObs.programId)               *>
+             newObs.asterismId.traverse(tryFindAsterism(t, _)) *>
+             newObs.targetId.traverse(tryFindTarget(t, _))     *>
```
swalker2m, line 127: Checking the validity of the `asterismId` or `targetId` is unrelated and should have been done before now anyway.

modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
```diff
@@ -195,20 +195,19 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, *]: Eq]
 
   def constructAndPublish[U <: T](
     cons: Tables => ValidatedInput[State[Tables, U]]
-  ): F[U] = {
-    val fu = EitherT(
+  ): F[ValidatedInput[U]] = {
```
swalker2m, line 198: So instead of throwing errors here we pass them up so they can be written in the query results.

toddburnside:
> I'm convinced by the `Payload` concept but on the fence about the error stuff.

One little thing about having the InputErrors in the payload is that it seems to introduce some inconsistencies with queries. I'm assuming that things like malformed IDs in the mutations would be reported in the InputErrors? Obviously this wouldn't be the case for queries.  I'm not sure if that's a big deal or not.

rpiaggio:
Nice. This will help detect the cause of errors.

However, the Apollo protocol for web sockets already provides a mechanism for the server to report errors to the client (the `error` message).  Could we still use that when we send mutations over a web socket? That way, we can have a single error handling mechanism in the client.

Alternatively, `data` messages support in their `"payload"` an optional `"error"` field besides `"data"`. In these cases the `data` message can be sent with all the `InputError` fields in the `"error"`. `clue` treats `data` messages with `"error"` present as errors, so we would still have a single mechanism.

swalker2m:
> One little thing about having the InputErrors in the payload is that it seems to introduce some inconsistencies with queries. I'm assuming that things like malformed IDs in the mutations would be reported in the InputErrors? Obviously this wouldn't be the case for queries.

A malformed ID actually violates the schema (ids are custom scalars) so the error would be reported in the same way as for a query: in the `errors` field (sibling of `data`) in the response.  The message itself is kind of unwieldy but you can get down to the actual problem if you read it

```
 "Variable '$createObservation' expected value of type 'CreateObservationInput!' but got: {\"programId\":\"x-2\",\"name\":\"Biff\",\"targetId\":\"t-2\"}. Reason: 'programId' Expected an ID of type p-(0|[1-9a-f][0-9a-f]*) 
```

Errors reported as part of the `Payload` union on the other hand are defined in the schema itself and so it affords the user an opportunity to do something about it.  In other words, errors "as data".  That said, I think as I have it here in the PR the possible errors are closer to programming errors than user errors and probably shouldn't be reported this way.  For example, referring to target `t-2` when that target doesn't exist.  An example of an error that would be more useful to a user would be choosing an account name that is already taken in a sign-up API.

Sometimes it isn't very clear cut though which errors rise to the level of something that should appear in the schema vs. programming errors that can be relegated to the `errors` field in the response.  I guess it's sort of like the difference between runtime exceptions and checked exceptions in Java.

swalker2m:
An example of the kind of error that probably does make sense in the schema from "Production Ready GraphQL":

```
type Mutation {
  signUp(email: string!, password: String!): SignUpPayload
}
union SignUpPayload =
  SignUpSuccess |
  UserNameTaken |
  PasswordTooWeak
```

rpiaggio:
Reading your example I see the need in some cases for structured error data that must be explicitly handled by the client. I see using the mechanism similar to defining a new custom exception vs just using good old `Exception`.

I found this article along the same lines, and which goes a bit further: https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/

The article suggests having an 
``` graphql
interface Error {
  message: String!
}
```
which seems like a good way to future-proof against new error types.

(Also, the article uses `*Result` which I find more semantic than `*Payload`).

All this makes me wish that GraphQL had first class `error` types that we could treat differently. For example: we could generate exception classes and actually `raiseError` when the result of executing a mutation returns one of these. We could actually convene that all types extending `Error` should be treated like this and have clue do that.

toddburnside:
> A malformed ID actually violates the schema (ids are custom scalars) so the error would be reported in the same way as for a query: in the `errors` field (sibling of `data`) in the response. The message itself is kind of unwieldy but you can get down to the actual problem if you read it

Ah, yeah. I guess I remember seeing that message - just didn't remember where it came from.

swalker2m:
I'll close this old payload PR that included error codes in the return type.  That part of the pattern is not as clear cut as just the idea of having a payload wrapper object.

PR #186, opened 2021-03-05
Observation Subject

swalker2m:
This PR not quite ready and lacks test cases and cleanup but I wanted to throw it out for discussion.  It would:

- Rename "observationTarget" to "subject".  I needed a good concise name for asterism or target and couldn't come up with a better name.
- Remove the target and asterism `shareWithObservations` mutations.
- Add a `setSubject` mutation that takes an observation list and a subject (asterism or target) and assigns the subject to all the observations.  (You can delete the subject by not defining either asterism or target.  I'll add an explicit `deleteSubject` to make this clearer.)
- Fix the events so that any targets that were heretofore quietly removed from an observation when the subject is changed generate an event as well

It would certainly be simpler and more consistent for the `setSubject` to work on just one observation instead of a list of observations but I did it this way to match the existing sharing methods.


modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -154,5 +165,51 @@ object ObservationRepo {
           Tables.constraintSetObservation,
           ConstraintSetModel.ConstraintSetEvent.updated
         )
+
+      override def setSubject(
+        oids:    List[Observation.Id],
+        subject: Option[Either[Asterism.Id, Target.Id]]
+      ): F[List[ObservationModel]] = {
```
swalker2m, line 172: This is a bit extra embarrassing.  I'll try to simplify before I mark it for final review.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -91,9 +90,7 @@ object ObservationModel extends ObservationOptics {
     observationId: Observation.Id,
     existence:     Input[Existence]   = Input.ignore,
     name:          Input[String]      = Input.ignore,
-    status:        Input[ObsStatus]   = Input.ignore,
-    asterismId:    Input[Asterism.Id] = Input.ignore,
-    targetId:      Input[Target.Id]   = Input.ignore
```
swalker2m, line 96: I'll remove the asterism and target id from the normal observation editor so that we don't have to support two different ways of doing this.  When you link a subject to an observation it has to generate extra events so it is best not to try to do it here.

rpiaggio:
> Rename "observationTarget" to "subject". I needed a good concise name for asterism or target and couldn't come up with a better name.

I am using `Aim` in Explore. If you don't like it, I'll switch to `Subject`, so that we use the same name.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
toddburnside: Doesn't allowing specifying a `subject` during ObservationCreation have the same issue? I think an edit event should be sent for that subject and it isn't right now.

swalker2m:
> I am using Aim in Explore. If you don't like it, I'll switch to Subject, so that we use the same name.

Oh, I didn't realize we had a term in the UI already for this.  I agree they should match.  Let's ask Andy/Bryan if they have a preference.

rpiaggio:
> Oh, I didn't realize we had a term in the UI already for this. I agree they should match. Let's ask Andy/Bryan if they have a preference.

I just came up with that a few days ago, I don't think it's merged. But I must say I'm starting to like `Subject` better.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: Yes, that's a good point.  I updated to send target/asterism events as appropriate on observation creation as well.  It's kind of convoluted code though :-/

toddburnside: Would it be better/easier to just not allow the subject to be set during creation? I don't think that would be an issue for Explore. Do you think it would be an issue for API users?

swalker2m: That was my first instinct and I debated it.  It is certainly easier and more consistent with general updating which doesn't allow it.  It would require two API calls to create an observation with a target instead of just one though and it seemed like a common use case would be to load a target list and then create observations for individual targets.

I don't know, I'm certainly willing to back it out.

toddburnside: Explore, at the moment at least, treats them as separate steps. I don't have a good idea of how people who work directly with the UI will be using it.

toddburnside:
> It would certainly be simpler and more consistent for the `setSubject` to work on just one observation instead of a list of observations but I did it this way to match the existing sharing methods.

Is there a need to match the existing sharing methods? I believe Explore always sends a list of one element at this point.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: It might not have made the doc, but we anticipated selecting a target and clicking the +Obs button to get a popup like this one:

Screen Shot 2021-03-08 at 14 52 21

which implies that it all happens in a transaction.  There would be similar pages in the constraints and configuration pages presumably.

toddburnside: Perhaps it is in the docs, but not yet in Explore.

toddburnside: If we include the subject in Create, it seems like we should also include it in Edit. And maybe we wouldn't need a separate Set Subject? Admittedly, if we included Subject, Contraints and Configuration in the Edit, the code for sending notifications would get complicated. So, I guess if we expect subsequent setting of subject/constraint/config to be done individually, having SetSubject, SetContraintSet and SetConfiguration is OK.

swalker2m: > If we include the subject in Create, it seems like we should also include it in Edit.

I believe it should be an option to set up these links during observation creation because it needs to be done as a single transaction to accommodate interactions like the "New" dialog I posted.  Once it is available there, I agree putting it in the editor as well seems natural, though I do think that thereafter you are editing the links individually.

The problem is it may be a bit challenging.  I'll give it a shot though.

swalker2m:
> Is there a need to match the existing sharing methods? I believe Explore always sends a list of one element at this point.

I suppose it would accommodate assigning a target or asterism to a group of observations in a single step.  Maybe it isn't critical though.  If we remove the `setSubject` mutation altogether in favor of doing it in the normal editor then I guess it goes away regardless.

rpiaggio:
Multi-drag is a desired feature. When we get around to it, it'd be nice to have a method that allows assigning a target/asterism to multiple observations at once.

swalker2m:
Still an untested draft, but now updated to:

* Rename "subject" to "pointing" as per suggestion.
* Allow specifying asterism / target in the editor (with appropriate events).

As it stands, the pointing can be set in observation create, edit, and via a separate update pointing that accepts multiple observation ids.  This would be inconsistent with all the other sharing.  If we keep these changes we should consider specifying asterism targets in its editor, program observations in its editor, etc. with appropriate events.

The code is admittedly kind of difficult but on the plus side I think assigning the pointing is more complicated than any of the other sharing options.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: The latest version again allows asterism/target in the editor.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -130,7 +133,36 @@ object ObservationModel extends ObservationOptics {
...
         a.status,
-        a.asterismId
+      )}
+
+  }
+
+  final case class EditPointing(
+    observationIds: List[Observation.Id],
+    asterismId:     Option[Asterism.Id],
+    targetId:       Option[Target.Id]
```
swalker2m, line 143: Instead of using `Input`s here I went with the simpler `Option`.  In the case of a general edit, we have to distinguish unsetting vs ignoring pointing.  Here presumably if you explicitly edit pointing then it can be assumed you don't want to ignore it.

swalker2m:
> As it stands, the pointing can be set in observation create, edit, and via a separate update pointing that accepts multiple observation ids. This would be inconsistent with all the other sharing. If we keep these changes we should consider specifying asterism targets in its editor, program observations in its editor, etc. with appropriate events.

I'll need to switch back to the observe API at this point so'll I'll mark this ready for review.  The questions in addition to normal code review are:

1. whether this seems like an acceptable model going forward for all the sharing cases (i.e., setting up links when creating, when editing in general, and then in a separate sharing mutation that accepts multiple ids as well)
2. whether we should wait until the other cases (asterism's targets, program's targets, etc.) have been updated before merging this one

toddburnside:
> 1. whether this seems like an acceptable model going forward for all the sharing cases (i.e., setting up links when creating, when editing in general, and then in a separate sharing mutation that accepts multiple ids as well)
> 
> 2. whether we should wait until the other cases (asterism's targets, program's targets, etc.) have been updated before merging this one

1. I think the model works. There is obviously overlap been the update/updatePointing, but it should make it easier for users of the API.
2. I don't see a need to wait until all the cases are updated. We do need to sync Explore with each change, but I prefer to not to let PRs get too stale.

toddburnside (approved):
LGTM

swalker2m:
> We do need to sync Explore with each change, but I prefer to not to let PRs get too stale.

I turned off automatic deployment for now.

toddburnside:
> I turned off automatic deployment for now.

After I get my constraint set tree PR in Explore through the gate, I can look at updating Explore for this PR.

rpiaggio:
> > 1. whether this seems like an acceptable model going forward for all the sharing cases (i.e., setting up links when creating, when editing in general, and then in a separate sharing mutation that accepts multiple ids as well)
> > 2. whether we should wait until the other cases (asterism's targets, program's targets, etc.) have been updated before merging this one
> 
> 1. I think the model works. There is obviously overlap been the update/updatePointing, but it should make it easier for users of the API.
> 2. I don't see a need to wait until all the cases are updated. We do need to sync Explore with each change, but I prefer to not to let PRs get too stale.

Agree on both points.

PR #192, opened 2021-03-12
Allows for unsetting the constraint set to observation link

toddburnside:
This is a temporary fix to make the constraint set UI work until the `observation`  <-> `pointing` api revamp has been completed (#186). At that point the API for `observation` <-> `constraint` set will  be reworked to match.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
@@ -3,13 +3,16 @@
 
 package lucuma.odb.api.repo
 
-import lucuma.odb.api.model.{ObservationModel, PlannedTimeSummaryModel}
+import lucuma.odb.api.model.{InputError, ObservationModel, PlannedTimeSummaryModel}
+import lucuma.odb.api.model.syntax.validatedinput._
 import lucuma.core.model.{Asterism, ConstraintSet, Observation, Program, Target}
 import lucuma.odb.api.model.ObservationModel.ObservationEvent
+import cats.MonadError
 import cats.effect.Sync
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import lucuma.odb.api.model.ConstraintSetModel
+import monocle.Lens
 
 sealed trait ObservationRepo[F[_]] extends TopLevelRepo[F, Observation.Id, ObservationModel] {
 
@@ -48,6 +51,8 @@ sealed trait ObservationRepo[F[_]] extends TopLevelRepo[F, Observation.Id, Obser
   def shareWithConstraintSet(oid: Observation.Id, csid: ConstraintSet.Id): F[ObservationModel]
 
   def unshareWithConstraintSet(oid: Observation.Id, csid: ConstraintSet.Id): F[ObservationModel]
+
+  def unsetConstraintSet(oid: Observation.Id)(implicit F: MonadError[F, Throwable]): F[ObservationModel]
 }
 
 object ObservationRepo {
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -154,5 +159,31 @@ object ObservationRepo {
           Tables.constraintSetObservation,
           ConstraintSetModel.ConstraintSetEvent.updated
         )
+
```
toddburnside, line 162: We need to be able to unset the link from explore without knowing the currently linked constraint set. It is possible to get the current link in explore, but the final API probably won't require it, so I'd like to avoid plumbing all that in only to rip it out.

modules/core/src/main/scala/lucuma/odb/api/repo/OneToManyUnique.scala
```diff
@@ -49,17 +49,15 @@ object OneToManyUnique {
     bToA:  SortedMap[B, A]
   ): OneToManyUnique[A, B] = new OneToManyUnique[A, B] {
 
+    // "Overwrites" existing links - there is no notification for the elements that are unlinked
```
toddburnside, line 52: This matches the current behavior for observation <-> pointing and doesn't allow for sending edit events for the constraint sets that are silently "unlinked". This will also be fixed in the rewrite.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -154,5 +159,31 @@ object ObservationRepo {
...
+        val doUpdate: F[(ObservationModel, Option[ConstraintSetModel])] = 
+          tablesRef.modify { oldTables => 
+            val obs = obsLens.get(oldTables).toValidNec(InputError.missingReference("id", oid.show))
+            val cs  = 
+              obs.fold(
+                _ => None,
+                o => oldTables.constraintSetObservation.selectLeft(o.id).flatMap(csId => Tables.constraintSet(csId).get(oldTables))
+              )
+            val tables = cs.fold(oldTables)(_ => Tables.constraintSetObservation.modify(_.removeRight(oid))(oldTables))
+            (tables, obs.map(o => (o, cs)))
```
swalker2m, line 179: I suppose you could `obs.tupleRight(cs)` there if you want.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -154,5 +159,31 @@ object ObservationRepo {
           Tables.constraintSetObservation,
           ConstraintSetModel.ConstraintSetEvent.updated
         )
+
+      override def unsetConstraintSet(oid: Observation.Id)(implicit F: MonadError[F, Throwable]): F[ObservationModel] = {
+        val focusObsLens = focusOn(oid)
+
+        val obsLens: Lens[Tables, Option[ObservationModel]] =
+          Lens[Tables, Option[ObservationModel]](focusObsLens.get)(oo => focusObsLens.set(oo))
```
swalker2m, line 167: Does this differ from `focusObsLens`?

swalker2m (approved):
Makes sense to me.  :+1:  Sorry the pointing/observation thing has dragged on so long!

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
toddburnside: Nope. I was having troubles getting everything to line up. I was sure I copied these optics from somewhere else and they worked, but I have no idea where. I'm probably imagining things...

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
toddburnside: Nice

toddburnside (commented):
> Sorry the pointing/observation thing has dragged on so long!

No worries. I have no shortage of things to work on. :)

PR #194, opened 2021-03-15
Nonempty string scalar

swalker2m:
Adds a `NonEmptyStringType` GraphQL scalar and converts the various names to `NonEmptyString` and `Option[NonEmptyString]`.  This moves emptiness checking from the realm of user input checks to schema violations.


rpiaggio (approved):
LGTM. Looking forward to see how it fits on the client side.

PR #218, opened 2021-04-01
Observe API Notes

swalker2m:
This PR seeks feedback on, as of yet, work-in-progress ideas on enough of the Observe API to execute a single observation.

### DIFF
--- a/docs/ObserveApiProposal.md
+++ b/docs/ObserveApiProposal.md
@@ -1,0 +1,782 @@
+# Observe API Notes
+
+This document sketches out a few ideas for the Observe execution API.  It focuses on the execution
+of a single observation, ignoring for now scheduling and timeline features, daytime calibration 
+queues, weather, suggesting where to split observations, etc.  For the initial step at transitioning
+Web Seqexec to GPP, the user should simply enter an observation id to load and Observe will execute
+it.
+
+Following GraphQL itself, there are query, mutation and subscription aspects to the API.
+
+
+## Query API
+
+Beginning with the query API, we'll introduce an `Execution` type to house runtime and execution
+related data for an observation (suitably access controlled).  The client accesses  it via an 
+observation id. Since the given observation may not exist, the result is optional.
+
+
+```
+type Query {
+  …
+  execution(
+    observationId: ObservationId!
+  ): Execution
+  …
+}
+```
+
+`Execution` can also (or perhaps alternatively) be returned from a field in the `Observation` type 
+itself.
+
+The first challenge for the API is that each instrument has its own configuration details and 
+features.  This implies that the API will have a combination of generic parts that apply to any 
+instrument along with instrument-specific parts.
+
+```
+type Execution {
+
+  # Generic, any instrument, section
+  executionState: ExecutionState!
+  timeAccounting: TimeAccounting!
+  remainingTime:  PlannedTime! 
+  
+  # Mix of generic and instrument-specific information
+  events: ExecutionEvents! 
+
+  # Instrument-specific
+  executionConfig: ExecutionConfiguration!
+    
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/ObserveApiProposal.md
```diff
@@ -0,0 +1,782 @@
...
+
+_The following SDL sketches out the general idea but specific details are likely to change._
+
+```
+
+# Individual step time can be divided into categories.  Time spent in each 
+# category will be grouped. 
+type StepTime {
+
+  # Time spent executing configuration changes (e.g., swapping filters)
```
jluhrs, line 134: ... or offsetting the telescope.
I'm just making the point that configuring does not apply to the instrument only.

docs/ObserveApiProposal.md
```diff
@@ -0,0 +1,782 @@
...
+
+}
+
+type GmosNorthStep {
+
+  stepId:     StepId!
+
+  # Breakpoints that are programmed into the sequence generation, or present in
+  # manual sequence definitions.
+  breakpoint: Breakpoint!
```
jluhrs, line 543: Is this a breakpoint that comes in the sequence definition? Should Observe stop the sequence before or after the step?
Are they related to the manual breakpoints set by the user while running the sequence in Observe? Can they be disabled/enabled by the Observe user?

docs/ObserveApiProposal.md
```diff
@@ -0,0 +1,782 @@
...
+  offset:    Offset!
+}
+
+```
+
+As discussed in the SDL comments above, it is expected that Observe will pull the next atom to
+execute at the beginning of the observation and each time thereafter that an atom is completed.
+This allows the course of the sequence to change over its execution according to conditions or
+other considerations. The `possibleFuture` shows what steps are anticipated for reference in the UI
+but should not be relied upon for actual execution.
```
jluhrs, line 612: This will be a mayor change in transforming Seqexec into Observe. The internal model must reflect this `currentAtom` + `posibleFuture`. I guess the UI should also show the `possibleFuture` in distinctive way, and should give some clue to the user when the `possibleFuture`  is replaced by a new `currentAtom`

docs/ObserveApiProposal.md
```diff
@@ -0,0 +1,782 @@
...
+  
+  # Science sequence describes how to configure the instrument to acheive its
+  # science goals. 
+  science:     ExecutionSequence!  
+  
+}
+
+# The execution sequence pairs the next atom that should be executed with a 
+# `possibleFuture` sequence for reference only.  Once the `nextAtom` is
+# complete, a new query is required to get the new `nextAtom`.
```
jluhrs, line 461: Could an Atom that Observe already loaded be changed from Explore? What if Observe already started executing it?

swalker2m: That's a good question.  It seems to me like we should prevent that in both the UI and API once execution has started.  The idea with atoms is that they be completed in their entirety or abandoned altogether.  Editing them while running seems anathema to that arrangement.

docs/ObserveApiProposal.md
swalker2m: Yes, a breakpoint in the sequence definition.  The idea was that there are sometimes natural points in a sequence (as defined by instrument scientists) where the observer should stop and do something.  One such case is during acquisition to stop and see whether the target is acquired.  Whether there there other cases I'm not sure.

My thought was that it signaled a stop _before_ the step so that there is a way to signal a pause before beginning.

I think these would be the same as manual breakpoints set by the user.  How that works with "possible future" I'll admit to being unsure.  If the possible future changes then the breakpoints you may have carefully set could disappear.

docs/ObserveApiProposal.md
swalker2m: Yes, that all sounds right to me.

PR #225, opened 2021-04-04
Modify constraint/observation sharing API to match Pointing.

toddburnside:


modules/core/src/main/scala/lucuma/odb/api/repo/AsterismRepo.scala
```diff
@@ -103,7 +104,9 @@ object AsterismRepo {
         tablesRef.get.map { t =>
           t.observations.get(oid).flatMap(_.pointing).collect {
             case Left(aid) => aid
-          }.flatMap(t.asterisms.get)
+          }
+          .flatMap(t.asterisms.get)
+          .filter(a => includeDeleted || a.isPresent)
```
swalker2m, line 109: Oops, thanks for fixing this.

modules/core/src/main/scala/lucuma/odb/api/repo/ConstraintSetRepo.scala
```diff
@@ -53,13 +47,15 @@ object ConstraintSetRepo {
...
-          }
-          .map(x => deletionFilter(includeDeleted)(x))
+        tablesRef.get.map{ t => 
+          val ocs = for {
+            o <- t.observations.get(oid)
+            csid <- o.constraintSetId
+            cs <- t.constraintSets.get(csid)
+          } yield cs
+
+          ocs.filter(cs => includeDeleted || cs.isPresent)
```
swalker2m, line 57: i wonder if it would work to put the filter in the `for` expression. 
```
  cs <- t.constraintSets.get(csid)
  if includeDeleted || cs.isPresent
} yield cs
```
No big deal but would save a line I suppose.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -99,9 +95,7 @@ object ObservationRepo {
         includeDeleted: Boolean
       ): F[ResultPage[ObservationModel]] =
 
-        selectPageFromIds(count, afterGid, includeDeleted) { tables =>
-          tables.constraintSetObservation.selectRight(csid)
-        }
+        selectPageFiltered(count, afterGid, includeDeleted) { _.constraintSetId.exists(_ == csid) }
```
swalker2m, line 98: Might as well go full `===`.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -230,55 +181,79 @@ object ObservationRepo {
...
+        
+        // Don't send a change event for the old id if the id Input is Ignore or the value is unchanged.
+        // Also don't send an event for the new id (if Assign) if it is always the same as the old id.
+        def ids4Event[A: Eq](oa: Option[A], ia: Input[A]): List[A] = (oa, ia) match {
+          case (_, Ignore)                => Nil
+          case (Some(a), Unassign)        => List(a)
+          case (None, Unassign)           => Nil
+          case (Some(aOld), Assign(aNew)) => if (aOld === aNew) Nil else List(aOld, aNew)
+          case (None, Assign(aNew))       => List(aNew)
+        }
```
swalker2m, line 201: Thanks, I think I was generating extra events before.

swalker2m (approved):
Looks great to me!  As for pointing, this stuff gets tricky but it's probably the most flexible for API users.  :+1:

modules/core/src/main/scala/lucuma/odb/api/repo/ConstraintSetRepo.scala
toddburnside: Good idea.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
toddburnside: Yeah, it was generating events for the current pointing on an edit event even if it wasn't being changed.

PR #241, opened 2021-04-20
Adds a `totalCount` field to the paging connection

swalker2m:
A `totalCount` field tells the caller, regardless of the page size, how many total items are forthcoming.  This is useful, for example, to determine whether a constraint set is shared by multiple observations.  You can select the observations associated with the constraint set and ask for its `totalCount`.

For example, to see the total target count for the `targets` query (along with the target names):

```
query Targets {
  targets(programId: "p-2", first: 100) {
    totalCount
    nodes {
      name
    }
  }
}
```

might return (say) 3 targets

```
{
  "data": {
    "targets": {
      "totalCount": 3,
      "nodes": [
        {
          "name": "NGC 5949"
        },
        {
          "name": "NGC 3269"
        },
        {
          "name": "NGC 3312"
        }
      ]
    }
  }
}
```

Asking for only 2 targets per page would return the two selected targets but the `totalCount` remains `3`:

```
query Targets {
  targets(programId: "p-2", first: 2) {
    totalCount
    nodes {
      name
    }
  }
}
```

```
{
  "data": {
    "targets": {
      "totalCount": 3,
      "nodes": [
        {
          "name": "NGC 5949"
        },
        {
          "name": "NGC 3269"
        }
      ]
    }
  }
}
```

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
@@ -120,18 +120,23 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, *]: Eq]
     includeDeleted: Boolean
   )(
     predicate: T => Boolean
-  ): F[ResultPage[T]] =
+  ): F[ResultPage[T]] = {
+
+    val isSelected: T => Boolean =
+      t => (includeDeleted || t.isPresent) && predicate(t)
 
     tablesRef.get.map { tables =>
-      val all    = mapLens.get(tables)
+      val all = mapLens.get(tables)
 
       page(
         count,
         afterGid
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/lucuma/odb/api/repo/TopLevelRepo.scala
```diff
@@ -120,46 +120,58 @@ abstract class TopLevelRepoBase[F[_]: Monad, I: Gid, T: TopLevelModel[I, *]: Eq]
...
-      val all    = mapLens.get(tables)
+      val all = mapLens.get(tables)
 
       page(
         count,
         afterGid
           .fold(all.valuesIterator)(gid => all.valuesIteratorFrom(gid).dropWhile(t => TopLevelModel[I, T].id(t) === gid))
-          .filter(t => (includeDeleted || t.isPresent) && predicate(t))
+          .filter(isSelected),
+        all.count { case (_, t) => isSelected(t) }
```
cquiroz, line 136: Nice

cquiroz (approved):
Looks good, it will be very useful

PR #291, opened 2021-05-27
Events Part 2

swalker2m:
Adds at least an initial version of mutations that Observe can use to add sequence, step, and dataset events.  Deletes the dataset creation input and the dataset repo in favor of generating datasets from the events themselves eliminating a potential source of inconsistency.  Executed steps can be culled from the event information as well, which will happen in a subsequent PR.


modules/core/src/main/scala/lucuma/odb/api/schema/ExecutionEventMutation.scala
```diff
@@ -0,0 +1,109 @@
...
+    List(
+      addSequenceEvent,
+      addStepEvent,
+      addDatasetEvent
+    )
+
+}
+
+object ExecutionEventMutation extends ExecutionEventMutation {
+
```
cquiroz, line 108: empty body?

swalker2m: Oops thanks, I'll take away the braces.

PR #295, opened 2021-05-31
No sharing experiment

swalker2m:
Experimental PR that makes constraint sets a property of observations.  Accordingly, it removes constraint set sharing and constraint set ids altogether.  Instead there is a new mutation, `updateConstraintSet`.  For example, this snippet will change the sky background to `GRAY` for observations `o-3` and `o-4`.

```
mutation BulkEditConstraints($bulkEditConstraints: BulkEditConstraintSetInput!) {
  updateConstraintSet(input: $bulkEditConstraints) {
    id
    constraintSet {
      skyBackground
    }
  }
}

"bulkEditConstraints": {
  "constraintSet": {
    "skyBackground": "GRAY"
  },
  "observationIds": [ "o-3", "o-4" ]
}
```

Next there is a query that groups observations by constraint set:

```
query ObservationsByConstraintSet {
  constraintGroups(first: 10) {
    nodes {
      constraintSet {
        cloudExtinction
        imageQuality
        skyBackground
        waterVapor
      }
      observationIds  # full observations also available
    }
  }
}
```

Like other queries that return a list of results, it supports paging.  This query is intended for the "Constraints" tab grouping.

While experimental, I think this cleans up and simplifies the code and API considerably.  If it is usable in the UI we should consider the same approach for configurations and targets.


modules/core/src/main/scala/lucuma/odb/api/schema/ConstraintSetSchema.scala
```diff
@@ -167,36 +132,57 @@ object ConstraintSetSchema {
...
+        Field(
+          name        = "observationIds",
+          fieldType   = ListType(ObservationIdType),
+          description = Some("IDs of observations that use the same constraints"),
+          resolve     = _.value.observationIds.toList
+        ),
+
+        Field(
+          name        = "observations",
+          fieldType   = ListType(ObservationType[F]),
```
swalker2m, line 153: I suppose this should be paged as well ...

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -265,29 +233,41 @@ object ObservationRepo {
...
       ): F[ObservationModel] =
 
         (edit.editor, edit.pointing).mapN { case (e, (a, t)) =>
-          doEdit(List(edit.observationId), e, a, t, edit.constraintSetId)
+          doEdit(List(edit.observationId), e, a, t)
         }.liftTo[F].flatten.map(_.head)
 
+
+      override def bulkEditConstraintSet(
+        edit: ConstraintSetModel.BulkEdit
```
swalker2m, line 250: One nice thing, I think, is that you can bulk edit the constraint sets of observations whether or not they share constraint sets in common.  For example, you could set the sky background to `GRAY` across all observations regardless of what constraints they currently hold.

swalker2m:
Example group query results:

```
{
  "data": {
    "constraintGroups": {
      "nodes": [
        {
          "constraintSet": {
            "cloudExtinction": "ONE_POINT_FIVE",
            "imageQuality": "TWO_POINT_ZERO",
            "skyBackground": "BRIGHT",
            "waterVapor": "WET"
          },
          "observationIds": [
            "o-2"
          ]
        },
        {
          "constraintSet": {
            "cloudExtinction": "THREE_POINT_ZERO",
            "imageQuality": "TWO_POINT_ZERO",
            "skyBackground": "GRAY",
            "waterVapor": "WET"
          },
          "observationIds": [
            "o-3",
            "o-4"
          ]
        }
      ]
    }
  }
}
```

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
cquiroz: Cool, I guess that leaves the other constraints as they are for those observations

andrewwstephens:
This looks great!
I am going to miss being able to name my constraint set with emojis, but I'll get over it.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
andrewwstephens: Now we need to support multi-select!

rpiaggio:
This is great. I'll start experimenting with this in the UI. I've deployed this branch to Heroku as a new development app `lucuma-odb-no-sharing` for that purpose.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
swalker2m: Is it ever the case that a human API-user will want to just update all observations in the program?  I could add an alternative to specify the program id instead of a list of observation ids if so.

rpiaggio:
I notice there's still a `name` in `ConstraintSet`. Will this be removed?

Also, it would help if we defined what we want for UI now in the Constraints View. I'm assuming that the tree on the left will show constraints sets grouped by same properties; and allow multiple selection?

swalker2m:
> I notice there's still a name in ConstraintSet. Will this be removed?

That's a good question.  I wasn't sure.  We could keep the name, in which case it opens the possibility of two sets of constraints being the same _except_ for the name.  So that might be confusing (though fixable) but it would make it possible to identify constraint sets by name.

swalker2m:
> I'm assuming that the tree on the left will show constraints sets grouped by same properties; and allow multiple selection?

I think it will look pretty similar to what we have now.  Unique constraint sets shown in the view on the left with the observations that use them nested beneath.  That corresponds to the `constraintGroups` query output, I think.

Multi-select for dragging observations seems like it would be useful but I don't know if it is necessary to multi-select constraints.  Dragging observations into a constraint set would cause a bulk edit of all selected observations to match the constraint set target.

swalker2m:
Switched to paging `observations` in the constraint set grouping.  For example here are two different ways to get the associated observation ids.  First the query:

```
query ObservationsByConstraintSet {
  constraintGroups(first: 10) {
    nodes {
      constraintSet {
        cloudExtinction
        imageQuality
        skyBackground
        waterVapor
      }
      observationIds
      observations(first: 10) {
        nodes {
          id
          name
        }
      }
    }
  }
}
```

And sample results:

```
{
  "data": {
    "constraintGroups": {
      "nodes": [
        {
          "constraintSet": {
            "cloudExtinction": "THREE_POINT_ZERO",
            "imageQuality": "TWO_POINT_ZERO",
            "skyBackground": "BRIGHT",
            "waterVapor": "WET"
          },
          "observationIds": [
            "o-2",
            "o-3",
            "o-4"
          ],
          "observations": {
            "nodes": [
              {
                "id": "o-2",
                "name": "NGC 5949"
              },
              {
                "id": "o-3",
                "name": "NGC 3312"
              },
              {
                "id": "o-4",
                "name": "Observation"
              }
            ]
          }
        }
      ]
    }
  }
}
```

toddburnside:
> > I notice there's still a name in ConstraintSet. Will this be removed?
> 
> That's a good question. I wasn't sure. We could keep the name, in which case it opens the possibility of two sets of constraints being the same _except_ for the name. So that might be confusing (though fixable) but it would make it possible to identify constraint sets by name.

It seems like allowing what are essentially non-independent items (they don't exist apart from an observation) to have names adds a lot of complications. If the constraints for an observation are edited to match an existing set of constraints except for the name, what do we show on the constraints tab? Do we prompt the user to consolidate and choose the name? Constraints are simple enough that maybe the summary will suffice to identify them? But, it seems to me that configurations are more complicated and a name might be helpful?

Just thinking out loud...

swalker2m:
>  If the constraints for an observation are edited to match an existing set of constraints except for the name, what do we show on the constraints tab

If we have a name, we should probably treat it like any other constraint set property.  So you'd see two different constraint sets.  It'd be on the user to consolidate them if desired, either by manually editing the name of one of them or else by dragging its observations to the other constraint set.

This brings up a complication in general.  If you edit a constraint set such that it matches another existing constraint set in other observations then the two groups are merged.  That would probably be disconcerting unless you can delay it until (say) focus is lost on the widget the user was interacting with.

andrewwstephens:
Constraint names seem redundant for simple constraint sets, however, they will be useful for non-obvious constraints like timing windows.  For example, a user might have a set of timing windows for when Io is not behind Jupiter, and this constraint set could be user-named something like "Io."  The timing windows for Europa would be in an otherwise identical constraint set named "Europa."

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
```diff
@@ -5,26 +5,23 @@ package lucuma.odb.api.model
...
 import monocle.macros.GenLens
 
+import scala.collection.immutable.SortedSet
+
 final case class ConstraintSetModel(
-  id:              ConstraintSet.Id,
-  existence:       Existence,
-  programId:       Program.Id,
-  name:            NonEmptyString,
+  name:            NonEmptyString,  // maybe we eliminate this?
```
swalker2m, line 24: Andy makes a good point about otherwise identical constraint sets that would be distinguished only on timing windows (e.g., Io vs Europa).  We could make the name optional though and perhaps most users would ignore it if we can default something based on CC, IQ, etc.

rpiaggio (approved):
LGTM

PR #298, opened 2021-06-02
Basic case porting I

cquiroz:
This PR is the first to get basic-case ported into lucuma-tmp-api. It defines a single query endpoint to get a list of confgurations for spectroscopy

This PR only defines the input types and also to get familiar with sangria and graphql, we still need to define the return type and hook it to the algorithm.

Sample of a query:
```
query{
  spectroscopy(input: {
    wavelength: {
      nanometers: 1000
    },
    simultaneousCoverage: {
      nanometers: 610
    },
    resolution: 2,
    signalToNoise: 10,
    spatialProfile: {
      sourceType: GAUSSIAN_SOURCE,
      fwhm: {
        arcseconds: 1
      }
    },
    spectralDistribution:BLACK_BODY,
    magnitude: {
      value: 3.55,
      band: J,
      system: VEGA
    },
    redshift: 1.5
  })
}	
```

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/SpatialProfileModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/SpatialProfileModel.scala
@@ -1,0 +1,81 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.api.model
+
+import io.circe.Decoder
+import io.circe.generic.semiauto._
+import lucuma.core.util.Enumerated
+import lucuma.core.util.Display
+
+object SpatialProfileModel {
+  final case class Input(sourceType: SpatialProfileType, fwhm: Option[GaussianSourceAngleInput]) extends Product with Serializable
+
+  sealed abstract class Units(
+    val angleUnit: AngleModel.Units
+  ) extends Product with Serializable
+
+  object Units {
+
+    case object Microarcseconds extends Units(AngleModel.Units.Microarcseconds)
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/SpatialProfileModel.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  }
+
+  final case class GaussianSourceAngleInput(
+    microarcseconds: Option[Long],
+    milliarcseconds: Option[BigDecimal],
+    arcseconds:      Option[BigDecimal],
+    fromLong:        Option[NumericUnits.LongInput[Units]],
+    fromDecimal:     Option[NumericUnits.DecimalInput[Units]]
+  )
```
swalker2m, line 46: Here is where I'd typically place something like

```scala
{
  def toAngle: ValidatedInput[Angle] =
    ValidatedInput.requireOne("fwhm",
      microarcseconds.map(Microarcseconds.readLong),
      milliarcseconds.map(Milliarcseconds.readDecimal),
      arcseconds.map(Arcseconds.readDecimal),
      fromLong.map(_.read),
      fromDecimal.map(_.read)
    )
}
```

swalker2m (approved):
Looks good to me. :+1:

modules/core/src/main/scala/lucuma/odb/api/model/SpatialProfileModel.scala
cquiroz: I guess that doesn't affect the graphql side right? You'd only use e.g. when executing the query?

swalker2m: Right, it would be used as part of validating the input and turning it into the model type you use internally for the matching.

PR #299, opened 2021-06-09
update observations query, make `first` parameter optional

swalker2m:
Updates the `observations` query to make `programId` optional and accept an optional `observationIds` list.  Now it works like:

* Only `programId` specified => Retrieve observations associated with the program
* Only `observationIds` specified => Retrieve matching observations (regardless of program)
* Neither specified => Retrieve all observations
* Both specified => Error

Updates `programs` query in a similar way, accepting an optional `programIds` list.  I would update the other top-level items similarly but it is unclear whether sharing with ids will live on.

Changes `first` parameter to be optional, defaulting for now to `Integer.MaxValue`.


swalker2m:
The plan would be to merge this and then rebase the no-sharing constraints branch.

swalker2m:
Screen Shot 2021-06-09 at 12 45 59

rpiaggio (approved):
LGTM

PR #302, opened 2021-06-16
Next Atom / Possible Future (Draft Observe API)

swalker2m:
An initial draft of the API Observe might use for executing an observation.  The idea is that it should pull the "next atom" one at a time from the database.  As it executes it sends sequence, step, and dataset events which influence the results of subsequent calls to get the next atom.  "Possible future" is available as well to show the expected future steps.

Screen Shot 2021-06-16 at 16 48 34

```
query NextAtom {
  observation(observationId: "o-3") {
    execution {
      executionConfig {
        instrument
        ... on GmosSouthExecutionConfig {
          science {
            nextAtom {
              steps {
                id
                stepType
                stepConfig {
                  ... on Science {
                    offset {
                      p { arcseconds }
                      q { arcseconds }
                    }
                  }
                  ... on Gcal {
                    continuum
                  }
                }
                instrumentConfig {
                  readout {
                    xBin
                    yBin
                  }
                  grating {
                    disperser
                  }
                  exposure {
                    seconds
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```


swalker2m:
Needs more testing but I wanted to get it out in case Javier wants to start looking at it sooner.

PR #305, opened 2021-06-24
active status

swalker2m:
Adds the observation "active" status corresponding to https://github.com/gemini-hlsw/lucuma-core/pull/349

I'm not happy with the name "active" for the field and mutation.  To set it to inactive you have to add, for example,

```
 "updateObservation": {
    "observationId": "o-2",
    "active": "INACTIVE"
  }
```

which seems weird.  Suggestions on a better name appreciated.


rpiaggio (approved):
LGTM

rpiaggio:
> I'm not happy with the name "active" for the field and mutation.

How about `activeStatus`?

swalker2m:
> How about `activeStatus`?

That's probably better.  Still, we'd have `status` (which is the observation status: for review, for activation, etc.) and `activeStatus`.

toddburnside:
> > How about `activeStatus`?
> 
> That's probably better. Still, we'd have `status` (which is the observation status: for review, for activation, etc.) and `activeStatus`.

`activeState`?

toddburnside (approved):
LGTM

swalker2m:
Renamed to `activeStatus`.

PR #308, opened 2021-06-29
CE3

swalker2m:
Updates to `cats-effect` 3.  To get there, I temporarily removed the dependency on `lucuma-sso` because it wants to include an incompatible version of `http4s`.  There is a stubbed out SSO client, which can hopefully be deleted in favor of the real client once `lucuma-sso` is on CE3 as well.

Removed and/or loosened constraints along the way where possible.  There were many superfluous requirements to supply `Effect` instances for example.

Much to my surprise it seems to work.


modules/core/src/main/scala/lucuma/odb/api/schema/ObservationSchema.scala
```diff
@@ -137,20 +139,20 @@ object ObservationSchema {
...
+          }
         ),
 
         Field(
           name        = "asterism",
           fieldType   = OptionType(AsterismType[F]),
           description = Some("The observation's asterism, if a multi-target observation"),
           arguments   = List(ArgumentIncludeDeleted),
-          resolve     = c => asterism[F](c).toIO.unsafeToFuture()
+          resolve     = c => c.unsafeToFuture(asterism[F](c))
```
cquiroz, line 155: Cool

modules/core/src/main/scala/lucuma/odb/api/schema/RepoContext.scala
```diff
@@ -86,31 +86,34 @@ final class RepoContextOps[F[_]: Effect](val self: Context[OdbRepo[F], _]) {
   def pagingTargetId: Either[InputError, Option[Target.Id]] =
     pagingGid[Target.Id]("TargetId")
 
-  def asterism[B](f: AsterismRepo[F] => F[B]): Future[B] =
-    f(self.ctx.asterism).toIO.unsafeToFuture()
+  def unsafeToFuture[B](fb: F[B])(implicit ev: Dispatcher[F]): Future[B] =
```
cquiroz, line 89: I guess this is the basis for your article right?

modules/core/src/test/scala/sangria/streaming/FS2IntegrationSpec.scala
```diff
@@ -19,70 +21,78 @@ import scala.concurrent.{Await, Future}
...
 
   test("failed") {
-    an [IllegalStateException] should be thrownBy res(impl.failed(new IllegalStateException("foo")))
+    an [IllegalStateException] should be thrownBy execStream(_.failed(new IllegalStateException("foo")))
   }
 
-  test("onComplete handles success") {
-    val stream  = Stream.empty
-    val count   = new AtomicInteger(0)
+  def execOnComplete(s: Stream[IO, _root_.fs2.INothing]): Assertion = {
```
cquiroz, line 71: `INothing`?

cquiroz (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/repo/EventService.scala
```diff
@@ -22,11 +21,7 @@ final class EventService[F[_]: Logger](
 )(implicit F: Concurrent[F]) {
 
   def subscribe: Stream[F, Event] =
-    topic.subscribeSize(100)
-      .evalTap { case (e, i) =>
-        Logger[F].info(s"subscription event ($i still queued): $e")
-      }
-      .map(_._1)
+    topic.subscribe(100)
```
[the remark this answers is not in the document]
swalker2m: `subscribeSize` went away and I didn't see a replacement.  when i added this logging during the spinning dots of death days, i suspected that events weren't going through but that didn't seem to be the case anyway

modules/core/src/test/scala/sangria/streaming/FS2IntegrationSpec.scala
swalker2m:
```
  /** Alias for `Nothing` which works better with type inference.
    */
  type INothing <: Nothing
```

These are the "valuess" produced by an empty `Stream`.  The test case here is for the `onComplete` method of the Sangria streaming interfacing so empty stream seemed fine.

PR #312, opened 2021-07-06
Switch to sbt-lucuma

cquiroz:
Mostly it needs us to update headers and remove some unused implicits


swalker2m (approved):
Thanks, looks good.  :+1:

PR #315, opened 2021-07-09
Update monocle

cquiroz:
Note I removed the use of `@Lenses` but only implemented the minimum to get this to build


build.sbt
```diff
@@ -14,10 +14,10 @@ val fs2Version                  = "3.0.6"
 val jawnVersion                 = "1.2.0"
 val kindProjectorVersion        = "0.13.0"
 val logbackVersion              = "1.2.3"
-val lucumaCoreVersion           = "0.10.1"
+val lucumaCoreVersion           = "0.0.0+1-4570bee1-SNAPSHOT"
```
swalker2m, line 17: Thanks this is great!  We have to wait for `lucuma-core` to be tagged and published though before merging this?

cquiroz: Maybe we can just do a lucuma release. If it is mayor we won't get auto merges

toddburnside (approved):
LGTM.

swalker2m (approved):
I'll add my :+1: to Todds.  Thanks again!

PR #317, opened 2021-07-12
Add science requirements to observations

cquiroz:
This is a first approximation at storing the science requirements params in the backend. I have some questions I want to ask inline but this is ready to be tested with the frontend

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -18,7 +18,7 @@ import eu.timepit.refined.types.string._
 import io.circe.Decoder
 import io.circe.generic.semiauto._
 import io.circe.refined._
-import monocle.{Lens, Optional}
+import monocle.{Focus, Lens, Optional}
 
 final case class ObservationModel(
   id:                   Observation.Id,
@@ -30,7 +30,8 @@ final case class ObservationModel(
   pointing:             Option[Either[Asterism.Id, Target.Id]],
   constraintSet:        ConstraintSetModel,
   plannedTimeSummary:   PlannedTimeSummaryModel,
-  config:               Option[InstrumentConfigModel.Reference]
+  config:               Option[InstrumentConfigModel.Reference],
+  scienceRequirements:  ScienceRequirements
 ) {
 
   def asterismId: Option[Asterism.Id] =
@@ -57,20 +58,22 @@ object ObservationModel extends ObservationOptics {
       o.pointing,
       o.constraintSet,
       o.plannedTimeSummary,
-      o.config
+      o.config,
+      o.scienceRequirements
     )}
 
 
   final case class Create(
-    observationId:   Option[Observation.Id],
-    programId:       Program.Id,
-    name:            Option[NonEmptyString],
-    asterismId:      Option[Asterism.Id],
-    targetId:        Option[Target.Id],
-    constraintSet:   Option[ConstraintSetModel.Create],
-    status:          Option[ObsStatus],
-    activeStatus:    Option[ObsActiveStatus],
-    config:          Option[InstrumentConfigModel.Create]
+    observationId:       Option[Observation.Id],
+    programId:           Program.Id,
+    name:                Option[NonEmptyString],
+    asterismId:          Option[Asterism.Id],
+    targetId:            Option[Target.Id],
+    constraintSet:       Option[ConstraintSetModel.Create],
+    status:              Option[ObsStatus],
+    activeStatus:        Option[ObsActiveStatus],
+    config:              Option[InstrumentConfigModel.Create],
+    scienceRequirements: Option[ScienceRequirementsModel.Create]
   ) {
 
     def create[F[_]: Monad, T](db: DatabaseState[T], s: PlannedTimeSummaryModel)(implicit S: Stateful[F, T]): F[ValidatedInput[ObservationModel]] =
@@ -89,7 +92,7 @@ object ObservationModel extends ObservationOptics {
 
         g <- config.traverse(_.create(db)).map(_.sequence)
 
-        o  = (i, p, pointing, constraintSet.traverse(_.create), g).mapN { (iʹ, _, pointingʹ, c, gʹ) =>
+        o  = (i, p, pointing, constraintSet.traverse(_.create), scienceRequirements.traverse(_.create), g).mapN { (iʹ, _, pointingʹ, c, q, gʹ) =>
           ObservationModel(
             iʹ,
             Present,
@@ -100,7 +103,8 @@ object ObservationModel extends ObservationOptics {
             pointingʹ,
             c.getOrElse(ConstraintSetModel.AnyConstraints),
             s,
-            gʹ.map(_.toReference)
+            gʹ.map(_.toReference),
+            q.getOrElse(ScienceRequirements.Default)
           )
         }
 
@@ -130,14 +134,15 @@ object ObservationModel extends ObservationOptics {
   }
 
   final case class Edit(
-    observationId:   Observation.Id,
-    existence:       Input[Existence]        = Input.ignore,
-    name:            Input[NonEmptyString]   = Input.ignore,
-    status:          Input[ObsStatus]        = Input.ignore,
-    activeStatus:    Input[ObsActiveStatus]  = Input.ignore,
-    asterismId:      Input[Asterism.Id]      = Input.ignore,
-    targetId:        Input[Target.Id]        = Input.ignore,
-    constraintSet:   Option[ConstraintSetModel.Edit] = None
+    observationId:       Observation.Id,
+    existence:           Input[Existence]                      = Input.ignore,
+    name:                Input[NonEmptyString]                 = Input.ignore,
+    status:              Input[ObsStatus]                      = Input.ignore,
+    activeStatus:        Input[ObsActiveStatus]                = Input.ignore,
+    asterismId:          Input[Asterism.Id]                    = Input.ignore,
+    targetId:            Input[Target.Id]                      = Input.ignore,
+    scienceRequirements: Option[ScienceRequirementsModel.Edit] = None,
+    constraintSet:       Option[ConstraintSetModel.Edit]       = None
   ) {
 
     def id: Observation.Id =
@@ -153,13 +158,17 @@ object ObservationModel extends ObservationOptics {
       (existence   .validateIsNotNull("existence"),
        status      .validateIsNotNull("status"),
        activeStatus.validateIsNotNull("active"),
+       scienceRequirements.traverse(_.editor),
        constraintSet.traverse(_.editor)
-      ).mapN { (e, s, a, c) =>
+      ).mapN { (e, s, a, r, c) =>
         for {
           _ <- ObservationModel.existence    := e
           _ <- ObservationModel.name         := name.toOptionOption
           _ <- ObservationModel.status       := s
           _ <- ObservationModel.activeStatus := a
+          _ <- State.modify[ObservationModel] { o =>
+            r.fold(o) { ed => ObservationModel.scienceRequirements.modify(ed.runS(_).value)(o) }
+          }
           _ <- State.modify[ObservationModel] { o =>
             c.fold(o) { ed => ObservationModel.constraintSet.modify(ed.runS(_).value)(o) }
           }
@@ -275,6 +284,9 @@ trait ObservationOptics { self: ObservationModel.type =>
     Optional[ObservationModel, Target.Id](_.pointing.flatMap(_.toOption)) { t =>
       _.copy(pointing = t.asRight[Asterism.Id].some)
     }
+
+  val scienceRequirements: Lens[ObservationModel, ScienceRequirements] =
+    Focus[ObservationModel](_.scienceRequirements)
 
   val constraintSet: Lens[ObservationModel, ConstraintSetModel] =
     Lens[ObservationModel, ConstraintSetModel](_.constraintSet)(a => _.copy(constraintSet = a))
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
@@ -49,13 +49,11 @@ object ScienceRequirementsModel {
 
   final case class Edit(
     mode:                     Input[ScienceMode]                                = Input.ignore,
-    spectroscopyRequirements: Input[SpectroscopyScienceRequirementsModel.Input] = Input.ignore
+    spectroscopyRequirements: Option[SpectroscopyScienceRequirementsModel.Input] = None
   ) {
 
     def editor: ValidatedInput[State[ScienceRequirements, Unit]] =
-      (mode.validateIsNotNull("mode"),
-       spectroscopyRequirements.validateNotNullable("scienceRequirements")(_.edit)
-       ).mapN { (m, s) =>
+      (mode.validateIsNotNull("mode"), spectroscopyRequirements.traverse(_.edit)).mapN { (m, s) =>
         for {
           _ <- ScienceRequirements.mode                     := m
           _ <- ScienceRequirements.spectroscopyRequirements := s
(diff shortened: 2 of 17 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
```diff
@@ -0,0 +1,70 @@
...
+import io.circe.generic.semiauto._
+import lucuma.core.optics.syntax.lens._
+import lucuma.odb.api.model.syntax.input._
+import clue.data.Input
+import monocle.Lens
+import monocle.Focus
+
+final case class ScienceRequirements(
+  mode:                     ScienceMode,
+  spectroscopyRequirements: SpectroscopyScienceRequirements
```
swalker2m, line 19: Nothing would prevent combining an `Imaging` mode with this?  One idea would be to make `ScienceRequirements` a `sealed trait` with a `mode` field and have `SpectroscopyRequirements` be an instance?   Or perhaps you want to store both imaging and spectroscopy requirements, if the user should enter both, and then use the `mode` to determine which one they really want?

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
```diff
@@ -0,0 +1,133 @@
...
+    val create: ValidatedInput[SpectroscopyScienceRequirements] =
+      (wavelength.traverse(_.toWavelength("wavelength")),
+       signalToNoiseAt.traverse(_.toWavelength("signalToNoiseAt")),
+       wavelengthRange.traverse(_.toWavelength("wavelengthRange")),
+       focalPlaneAngle.traverse(_.toAngle)
+      ).mapN { (cw, signalToNoiseAt, wavelengthRange, focalPlaneAngle) =>
+        SpectroscopyScienceRequirements(cw, resolution, signalToNoise, signalToNoiseAt, wavelengthRange, focalPlane, focalPlaneAngle, capabilities)
+      }
+    val edit: ValidatedInput[SpectroscopyScienceRequirements] =
+      create
```
swalker2m, line 70: When editing, you don't want to provide a way to remove an option once specified?

swalker2m (approved):
This is looking great!

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
cquiroz: I want to store both. This is an UI factor as I want to store the parameters in either case in case you switch.
Let's say you're in spectroscopy and change some params, then look at imaging and then return
Ideally we'd preserve the previously changed params.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
cquiroz: This was one of my questions. I'm not sure how this goes? Is the idea to set `null` to remove an option?
If so, yes we want to have that option

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
swalker2m: That makes sense.  :+1:

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  }
+
+  final case class Edit(
+    mode:                     Input[ScienceMode]                                = Input.ignore,
+    spectroscopyRequirements: Input[SpectroscopyScienceRequirementsModel.Input] = Input.ignore
+  ) {
+
+    def editor: ValidatedInput[State[ScienceRequirements, Unit]] =
+      (mode.validateIsNotNull("mode"),
```
cquiroz, line 56: I couldn't find how to make this optional. We may want to let the frontend edit without specifying the mode or the requirements.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
```diff
@@ -0,0 +1,133 @@
...
+     x.wavelengthRange,
+     x.focalPlane,
+     x.focalPlaneAngle,
+     x.capabilities
+    )
+  )
+}
+
+object SpectroscopyScienceRequirementsModel {
+  final case class Input(
```
cquiroz, line 51: I used this for Edit and Create. I'm not sure if this is correct. It seems that both patterns are used.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
```diff
@@ -0,0 +1,133 @@
...
+    milliarcseconds: Option[BigDecimal],
+    arcseconds:      Option[BigDecimal]
+  ) {
+    import Units._
+
+    val toAngle: ValidatedInput[Angle] =
+      ValidatedInput.requireOne("focalPlaneAngle",
+        microarcseconds.map(Microarcseconds.readLong),
+        milliarcseconds.map(Milliarcseconds.readDecimal),
+        arcseconds     .map(Arcseconds.readDecimal)
```
cquiroz, line 90: I didn't provide a `fromLong` and `fromDecimal` as it seems you can't really tell the units for decimals. Or do we want those methods?

modules/core/src/main/scala/lucuma/odb/api/schema/ScienceRequirementsSchema.scala
```diff
@@ -0,0 +1,157 @@
...
+    )
+
+  def ScienceRequirementsType[F[_]]: ObjectType[OdbRepo[F], ScienceRequirements] =
+    ObjectType(
+      name     = "ScienceRequirements",
+      fieldsFn = () =>
+        fields(
+          Field(
+            name        = "mode",
+            fieldType   = EnumTypeScienceMode,
```
cquiroz, line 143: Should these be `OptionType`?

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
swalker2m: You want to make it possible to not specify the mode?  If so, in the model it should be an `Option` and you'd `validateNullable` here.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
swalker2m: "It depends." Often when you create something there are a lot of fields that are required.  If the user doesn't specify one or more required fields then you can't really create the `Foo` object.  When you edit though, the idea is to only specify the things you want to change.  Other stuff that goes unspecified in the editor is left untouched.

In this case, everything is optional upon create so this may be fine.  On the other hand, it leaves us with no way to distinguish between wanting to remove something and wanting to leave it untouched.  Say for example:

```
  "capabilities" : null
```

in the editor to remove them.  For that you need a separate editor and the field becomes a nullable clue `Input`.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
swalker2m: See, for example, `FiniteDurationModel.Input.fromLong` and `fromDecimal`.  You specify the units along with the value.  It allows the user to use long or decimal values regardless of the units.

That said, those may be overkill.  Maybe we should ask Andy whether he finds it at all useful.  If not, we should remove them everywhere.  If so, we should add it here.

modules/core/src/main/scala/lucuma/odb/api/schema/ScienceRequirementsSchema.scala
swalker2m: Only if it is indeed optional in the model itself.  Here, as currently defined, the mode is always present so it will always be present if this field is included in the query.

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
cquiroz: mode can't really be an Option, I guess I'd have to pass it always. On the other hand the specific mode params can be optional.
I'm not sure why `validateNullable` takes a function as a param.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
cquiroz: I see, it seems we want to support nulling a field

modules/core/src/main/scala/lucuma/odb/api/model/ScienceRequirements.scala
swalker2m: If it is nullable (in that you can specify `null` and remove the optional field), and there is nothing to validate if it is specified, then you can use `toOptionOption` down in the state editor.  See `AsterismModel.Edit` and the `name` field.  It is nullable but there's nothing to validate.

swalker2m: I think here for the `mode` you can't unset it, you can only change the value (?). In that case it makes sense to me that it is `validateIsNotNull`.

cquiroz:
Here are some sample queries

Reading:
```graphql
query{
  observations(programId: "p-2") {
    nodes {
      id
      scienceRequirements {
        mode
        spectroscopyRequirements {
          wavelength {
            micrometers
          }
          resolution
          signalToNoise
          signalToNoiseAt {
            micrometers
          }
          wavelengthRange {
            micrometers
          }
          focalPlane
          focalPlaneAngle {
            arcseconds
          }
          capabilities
        }
      }
      
    }
  }
}
```

Response:
```graphql
{
  "data": {
    "observations": {
      "nodes": [
        {
          "id": "o-2",
          "scienceRequirements": {
            "mode": "IMAGING",
            "spectroscopyRequirements": {
              "wavelength": {
                "micrometers": 0.8
              },
              "resolution": 110,
              "signalToNoise": 4.5,
              "signalToNoiseAt": {
                "micrometers": 0.6
              },
              "wavelengthRange": {
                "micrometers": 0.1
              },
              "focalPlane": "IFU",
              "focalPlaneAngle": {
                "arcseconds": 0.00001
              },
              "capabilities": "NOD_AND_SHUFFLE"
            }
          }
        },
...
```

cquiroz:
And to edit:

```graphql
mutation {
  updateObservation(input: {
    observationId: "o-2",
    scienceRequirements: {
      spectroscopyRequirements: {
        wavelength: {
          micrometers: 0.8
        }
        resolution: 110
        signalToNoise: 4.5
        signalToNoiseAt: {
          micrometers: 0.6
        }
        wavelengthRange: {
          micrometers: 0.1
        }
        focalPlane: IFU
        focalPlaneAngle: {
          microarcseconds: 15
        }
        capabilities: NOD_AND_SHUFFLE
      }
    }
    
  })
...
```

cquiroz:
Fixed the ability to remove fields and added mutation for bulk edit

swalker2m:
Sounds great, I'll take a look at it in the morning.

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
```diff
@@ -0,0 +1,210 @@
...
+
+final case class SpectroscopyScienceRequirements(
+  wavelength:       Option[Wavelength],
+  resolution:       Option[PosInt],
+  signalToNoise:    Option[PosBigDecimal],
+  signalToNoiseAt:  Option[Wavelength],
+  wavelengthRange:  Option[Wavelength],
+  focalPlane:       Option[FocalPlane],
+  focalPlaneAngle:  Option[Angle],
+  capabilities:     Option[SpectroscopyCapabilities]
```
swalker2m, line 37: Ignorance revealing question: can you do multiple capabilities in the same observation?

swalker2m (approved):
I took another look and it all looks great. I'm sorry you had to wade through this!  I'll second my prior approval.  :+1:

modules/core/src/main/scala/lucuma/odb/api/model/SpectroscopyRequirements.scala
cquiroz: Not as specified now, We could of course make this a list for the eventuality?

swalker2m: Nah, we can change it later if need be.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -153,13 +158,17 @@ object ObservationModel extends ObservationOptics {
...
-      ).mapN { (e, s, a, c) =>
+      ).mapN { (e, s, a, r, c) =>
         for {
           _ <- ObservationModel.existence    := e
           _ <- ObservationModel.name         := name.toOptionOption
           _ <- ObservationModel.status       := s
           _ <- ObservationModel.activeStatus := a
+          _ <- State.modify[ObservationModel] { o =>
+            r.fold(o) { ed => ObservationModel.scienceRequirements.modify(ed.runS(_).value)(o) }
+          }
```
swalker2m, line 171: I wrote the similar case for `constraintSet` below but it bothered me that it was a bit awkward.  I'm sure there is a better way to do this.  One small improvement though would be:

```scala
          _ <- r.fold(State.get[ObservationModel].void) { ed =>
            ObservationModel.scienceRequirements.mod_(ed.runS(_).value)
          }
```

You have to import `import lucuma.core.optics.state.all._` for that.

cquiroz:
This is ready for a final review, it is made to match the PR for the frontend

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
```diff
@@ -297,6 +299,23 @@ object ObservationRepo {
...
 
+      override def bulkEditScienceRequirements(edit: ScienceRequirementsModel.BulkEdit): F[List[ObservationModel]] = {
+        val update =
+          for {
+            vos <- TableState.observation.lookupAllValidated[State[Tables, *]](edit.observationIds)
+            os  <- (vos, edit.scienceRequirements.editor).mapN { (os, ed) =>
+                     val os2 = os.map {
+                       ObservationModel.scienceRequirements.modify { cs => ed.runS(cs).value }
+                     }
+                     Tables.observations.mod(_ ++ os2.map(o => o.id -> o)).as(os2)
```
swalker2m, line 310: For what it's worth, which is probably just about nothing, `os2.fproductLeft(_.id)` is another option

modules/core/src/main/scala/lucuma/odb/api/schema/RefinedSchema.scala
```diff
@@ -4,8 +4,9 @@
 package lucuma.odb.api.schema
 
 import cats.syntax.all._
+import eu.timepit.refined.types.all.PosBigDecimal
 import eu.timepit.refined.types.all.PosInt
-import sangria.schema.{InputType, IntType, ScalarAlias}
+import sangria.schema.{InputType, IntType, BigDecimalType, ScalarAlias, ScalarType}
 import sangria.validation.ValueCoercionViolation
 
 trait RefinedSchema {
```
swalker2m, line 12: Maybe `NonEmptyString` in `GeneralSchema` belongs here too.

swalker2m (approved):
Looks great to me.

modules/core/src/main/scala/lucuma/odb/api/repo/ObservationRepo.scala
cquiroz: Right, will fix that

modules/core/src/main/scala/lucuma/odb/api/schema/RefinedSchema.scala
cquiroz: Indeed, I didn't want to move it but I thought the same

PR #318, opened 2021-07-12
Magnitude list editor

swalker2m:
Updates target magnitude editing to work like the proposed changes for no-sharing targets.  In particular, there is one `magnitudes` field in the `EditSiderealTargetInput`.  If included in the edit, the user must choose between `replaceList` and `editList`.  If `replaceList` then the list of magnitudes that you want is fully specified.  If `editList` then each item in the list operates on a particular magnitude entry either adding a new one, deleting an existing one, or editing an existing one.

Screen Shot 2021-07-12 at 18 08 59

The editor doesn't generate errors in some cases that you might (or might not) expect.  For example, if you try to delete a magnitude that you don't have in your list.  The idea was that it would allow you to reuse the editor for multiple targets that may or may not have said magnitude.  We can adjust this as necessary.


swalker2m:
This would, unfortunately, require some Explore change if merged.

swalker2m:
Example edit

```json
"magnitudes": {
  "editList":  [
   {
      "add":  { 
        "band": "H",
        "value": 19
      }
    },
    {
      "delete": "J"
    },
    {
      "edit": {
        "band": "K",
        "system": "VEGA"
      }
    }
  ]
}


```

toddburnside:
Is the current implementation the equivalent of `replaceList`?

swalker2m:
> Is the current implementation the equivalent of `replaceList`?

There is a `magnitudes` entry that is the equivalent of `"magnitudes": { "replaceList" : [ ... ] }` and there are also `modifyMagnitudes` and `deleteMagnitudes` entries which would be removed by this PR.

Screen Shot 2021-07-12 at 19 22 24

toddburnside (approved):
LGTM

PR #322, opened 2021-07-21
generalize grouping

swalker2m:
We will be adding grouping of observations by various criteria.  The existing grouping-by-constraint-set code was very specific to constraint sets so I generalized it to make it available for reuse.


toddburnside (approved):
LGTM

PR #323, opened 2021-07-21
generalize bulk edit

swalker2m:
Removes a bit of code duplication but more importantly makes it easier to add new observation bulk edit operations.


cquiroz (approved):
LGTM, thanks

PR #325, opened 2021-07-22
Group by science requirements

swalker2m:
Adds grouping by common science requirements.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -9,6 +9,7 @@ import lucuma.core.`enum`.{ObsActiveStatus, ObsStatus}
 import lucuma.core.optics.syntax.lens._
 import lucuma.core.model.{Asterism, Observation, Program, Target}
 import cats.{Eq, Monad}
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -9,6 +9,7 @@ import lucuma.core.`enum`.{ObsActiveStatus, ObsStatus}
 import lucuma.core.optics.syntax.lens._
 import lucuma.core.model.{Asterism, Observation, Program, Target}
 import cats.{Eq, Monad}
+import cats.Order.catsKernelOrderingForOrder
```
swalker2m, line 12: I always need this but I'm never sure what the right way to import it is.

cquiroz: Maybe cats.implicits.order?

toddburnside: You can use `cats.implicits._` or `cats.implicits.catsKernelOrderingForOrder` I believe.

toddburnside (approved):
LGTM

PR #330, opened 2021-07-30
Support adding and editing the base configuration of an observation

cquiroz:
For details see `docs/ScienceConfiguration.md`
We way want to add an `editScienceConfiguration` operation as well

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -4,9 +4,11 @@
 package lucuma.odb.api.model
 
 import lucuma.odb.api.model.Existence._
+import lucuma.odb.api.model.ScienceConfigurationModel.ScienceConfigurationModelEdit
 import lucuma.odb.api.model.syntax.input._
 import lucuma.core.`enum`.{ObsActiveStatus, ObsStatus}
 import lucuma.core.optics.syntax.lens._
+import lucuma.core.optics.state.all._
 import lucuma.core.model.{Asterism, Observation, Program, Target}
 import cats.{Eq, Monad}
 import cats.Order.catsKernelOrderingForOrder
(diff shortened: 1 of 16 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -34,7 +36,8 @@ final case class ObservationModel(
   constraintSet:        ConstraintSetModel,
   scienceRequirements:  ScienceRequirements,
   plannedTimeSummary:   PlannedTimeSummaryModel,
-  config:               Option[InstrumentConfigModel.Reference]
+  config:               Option[InstrumentConfigModel.Reference],
```
swalker2m, line 39: Maybe let's keep `config` last?  At some point I may want to figure out a new name.  Maybe `manualConfig`.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -140,44 +146,56 @@ object ObservationModel extends ObservationOptics {
...
+    val pointing: ValidatedInput[(Input[Asterism.Id], Input[Target.Id])] =
       (asterismId, targetId) match {
         case (Assign(_), Assign(_)) => InputError.fromMessage(s"Cannot assign both an asterism and a target to the observation").invalidNec
         case _                      => (asterismId, targetId).validNec
       }
 
+    def editOrCreateSciConfig(ed: Option[Option[Either[ScienceConfigurationModel, State[ScienceConfigurationModel, Unit]]]])(m: Option[ScienceConfigurationModel]): Option[ScienceConfigurationModel] =
+      ed.flatMap {
+        case Some(Left(s))  => s.some
+        case Some(Right(s)) => m.map(s.runS(_).value)
```
swalker2m, line 173: If you choose to edit but the science configuration doesn't exist yet, then essentially nothing happens correct?

swalker2m (approved):
Well it all looks perfectly cromulent to me.  Good work figuring out this mess! :+1:

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: But no need to change it now, I can move it later.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
cquiroz: Right, It is ignored, as we have discussed an alternative would be for it to fail in that case

swalker2m: Let's go with this as I'm not sure what the "right" answer is anyway.

PR #336, opened 2021-08-03
add some sanity tests

tpolecat:
I'm in the process of factoring out the http4s routing logic so it can be reused in SSO and other services, so I wanted to get some tests in here beforehand. I will follow up with another set for subscriptions. Thanks @swalker2m for the example queries.


swalker2m:
Just a note, anything target-related will be undergoing violent changes in the near future if all goes well.

PR #338, opened 2021-08-04
No sharing targets impl

swalker2m:
A preliminary no-sharing of targets update.  (I want to add more tests and consider what API changes may be needed.)

This PR:

* In the observation model, replaces asterisms and the existing `pointing: Option[Either[Asterism.Id, Target.Id]]` with `targets: TargetEnvironmentModel`.  The target environment currently houses an optional explicit base position and the collection of science targets.  In the future, guide stars, blind offset stars, etc would be found here as well.
* Makes it possible for any observation to have multiple targets (i.e., an asterism).
* Removes target ids, using the name for identifying targets to edit.
* Adds several grouping options: by individual targets, by the collection of science targets, and by the target environment as a whole.
* Updates bulk editing with `select` and `edit` fields and adds bulk editors for individual targets, the collection of science targets, and the target environment as a whole.

One potentially controversial update is to make a dedicated target renaming mutation.

```
      mutation EditTargetName($editTargetName: BulkEditTargetNameInput!) {
        updateScienceTargetName(input: $editTargetName) {
          id
          targets {
            science {
              name
            }
          }
        }
      }
```
```
      {
        "editTargetName": {
          "select": {
            "observationIds": [ "o-2" ]
          },
          "edit": {
            "oldName": "NGC 5949",
            "newName": "NGC 5949x"
          }
        }
      }
```
producing
```
      {
        "updateScienceTargetName": [
          {
            "id": "o-2",
            "targets": {
              "science": [
                {
                  "name": "NGC 5949x"
                }
              ]
            }
          }
        ]
      }
```
This simplified the other target editors because the name is only used there to identify targets.


modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
```diff
@@ -347,29 +338,34 @@ trait ObservationOptics { self: ObservationModel.type =>
   val activeStatus: Lens[ObservationModel, ObsActiveStatus] =
     Lens[ObservationModel, ObsActiveStatus](_.activeStatus)(a => _.copy(activeStatus = a))
 
-  val pointing: Lens[ObservationModel, Option[Either[Asterism.Id, Target.Id]]] =
-    Lens[ObservationModel, Option[Either[Asterism.Id, Target.Id]]](_.pointing)(a => _.copy(pointing = a))
+  val targets: Lens[ObservationModel, TargetEnvironmentModel] =
+    Lens[ObservationModel, TargetEnvironmentModel](_.targets)(a => _.copy(targets = a))
```
cquiroz, line 342: Would it make sense to use the `Focus` macro?

modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
```diff
@@ -58,56 +44,103 @@ object TargetModel extends TargetOptics {
...
+      listName:      String,
+      observationId: Observation.Id
+    ): ValidatedInput[TargetMap] = {
+
+      def alreadyExists: InputError =
+        InputError.fromMessage(
+          s"Cannot create a new ${listName.capitalize} target with name '$name' because one already exists in observation ${Gid[Observation.Id].show(observationId)}"
+        )
+
+      (toGemTarget, targetMap.get(name).as(alreadyExists).toInvalidNec(targetMap)).mapN { (t, m) =>
```
cquiroz, line 113: I guess there should be a way  to delete a target right?

swalker2m: That's in the `EditTargetAction` with its `add`, `delete`, and `edit` options.  Maybe it needs a better name now that I think about it.

modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
swalker2m: Thanks, I'll switch.  In general I prefer more explicit code even at the cost of boilerplate but `Focus` (which is just `GenLens`?) seems fine.

modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
```diff
@@ -228,69 +246,124 @@ object TargetModel extends TargetOptics {
...
-    existence: Option[Existence],
-    name:      Option[NonEmptyString],
-    key:       Option[EphemerisKey],
+  /**
+   * Input required to edit either a non-sidereal or sidereal target.
+   */
+  final case class Edit(
+    nonsidereal: Option[EditNonsidereal],
+    sidereal:    Option[EditSidereal],
+    rename:      Option[EditName]
```
cquiroz, line 255: Will this apply to either sideral/non-sidereal?

modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
```diff
@@ -331,72 +402,192 @@ object TargetModel extends TargetOptics {
...
+  val radialVelocity: Optional[Target, Option[RadialVelocity]] =
     siderealTracking.andThen(SiderealTracking.radialVelocity)
 
-  val parallax: Optional[TargetModel, Option[Parallax]] =
+  val parallax: Optional[Target, Option[Parallax]] =
     siderealTracking.andThen(SiderealTracking.parallax)
 
-  val magnitudes: Lens[TargetModel, SortedMap[MagnitudeBand, Magnitude]] =
-    lucumaTarget.andThen(Target.magnitudes)
+  val magnitudes: Lens[Target, SortedMap[MagnitudeBand, Magnitude]] =
```
cquiroz, line 590: Is this a circular reference?

modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
swalker2m: Yes, `rename` will execute this state editor: `lucuma.core.model.Target.name.assign_(newName)` in the end.  `Target` name is the same for all targets regardless of their `track`ing.

cquiroz (approved):
Looks great, I'm looking forward to somebody to do the UI side :smile:

modules/core/src/main/scala/lucuma/odb/api/model/TargetModel.scala
swalker2m: Actually now it's just a reference to `Target.magnitudes`. `TargetModel` the case class went away because without the id it offered nothing that `lucuma.core.model.Target` didn't.

toddburnside:
> {
        "editTargetName": {
          "select": {
            "observationIds": [ "o-2" ]
          },
          "edit": {
            "oldName": "NGC 5949",
            "newName": "NGC 5949x"
          }
        }
      }

Why do we need a dedicated mutation for renaming? I thought `name` was going to be part of the `select`.

swalker2m:
> Why do we need a dedicated mutation for renaming? I thought name was going to be part of the select.

I did start down that path but then I realized it would only work for editing single targets.  Once you edit a pair or more of targets then the specific edits have to be associated with the corresponding target names.  Making the name editable in the normal sidereal and non-sidereal editors and using them for identifying targets seemed too confusing.  A special purpose rename edit solved the problem and, I think, renaming will be uncommon.

toddburnside:
> I did start down that path but then I realized it would only work for editing single targets. Once you edit a pair or more of targets then the specific edits have to be associated with the corresponding target names. Making the name editable in the normal sidereal and non-sidereal editors and using them for identifying targets seemed too confusing. A special purpose rename edit solved the problem and, I think, renaming will be uncommon.

It seems to me that in the `editTargetName` mutation, the `oldName` is being used as a selector and could just go in the `select` object. I, personally, would find that less confusing than having an `oldName` and a `newName` in an object called `edit`.

Also, looking at the API in playground for the review app, it looks like the `updateSiderealScienceTarget` mutation only has ProgramId and a list of ObservationId as the selector. Unless I'm missing something, that means that when you edit, you are editing ALL of the targets for each specified observation (or all for the program). I think we need to be able to target more selectively. My understanding is that in the UI we will be grouping by individual targets, with the list of observations that contain those targets. So, we would need to be able to specify a list of observation ids AND the target name being used as a key. I was thinking the `select` would be ProgramId, list of ObservationId (optional) and target name (also optional).

modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
```diff
@@ -164,14 +164,36 @@ trait TargetMutation extends TargetScalars {
...
+
+  implicit val InputObjectTypeTargetEditSidereal: InputObjectType[TargetModel.EditSidereal] =
     deriveInputObjectType[TargetModel.EditSidereal](
       InputObjectTypeName("EditSiderealInput"),
       InputObjectTypeDescription("Sidereal target edit parameters"),
 
       DocumentInputField("magnitudes",    "Edit magnitudes"                                               ),
 
-      ReplaceInputField("existence",      EnumTypeExistence        .notNullableField("existence"  )),
       ReplaceInputField("name",           StringType               .notNullableField("name"       )),
```
cquiroz, line 197: Should this be a `NonEmptyString`?

modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
```diff
@@ -164,14 +164,36 @@ trait TargetMutation extends TargetScalars {
...
+      "Target creation parameters.  Choose 'nonSidereal' or 'sidereal'."
+    )
+
+
+  implicit val InputObjectTypeEditNonsidereal: InputObjectType[TargetModel.EditNonsidereal] =
+    deriveInputObjectType[TargetModel.EditNonsidereal](
+      InputObjectTypeName("EditNonsiderealInput"),
+      InputObjectTypeDescription("Nonsidereal target edit parameters"),
+
+      ReplaceInputField("name", StringType      .notNullableField("name")),
```
cquiroz, line 185: Should this be a `NonEmptyString`?

toddburnside:
Ah. I think I see what your intentions are with `updateSiderealScienceTarget` and `updateAllScienceTargets`. I'll put together some comments on that in a bit.

swalker2m:
A current multiple target mutation example.  There are a couple of different levels at which you can operate.  This one will work on an observations science target list.

It's a bit cumbersome to be honest.  If we want to support this kind of thing though then a single target name cannot be in the selector.  The selector just picks which observations the editor will work on.

```
mutation BulkEditScienceTargets($bulkEditScienceTargets: BulkEditScienceInput!) {
  updateAllScienceTargets(input: $bulkEditScienceTargets) {
    targets {
      science {
        name
        tracking {
          ... on Sidereal {
            coordinates {
              ra { hms }
              dec { dms }
            }
          }
        }
      }
    } 
  }
}
```

variable:

```
{
  "bulkEditScienceTargets": {
    "select": {
      "observationIds": [ "o-2" ]
    },
    "edit": {
      "editList": [
        {
          "edit": {
            "rename": {
              "oldName": "Bif",
              "newName": "Bid"
            }
          }
        },
        {
          "add": {
            "sidereal": {
              "name": "Zero",
              "ra": {
                "hms": "00:00:00.00"
              },
             "dec":{
                "dms": "00:00:00.00"
              }
            }
          }
        },
        {
          "edit": {
            "sidereal": {
              "name": "NGC 5949",
              "ra": {
                "hms": "11:11:11.11"
              }
            }
          }
        }
      ]
    }
  }
}
```

swalker2m:
Editing a single sidereal target:

```
mutation BulkEditTarget($bulkEditTarget: BulkEditSiderealInput!) {
  updateSiderealScienceTarget(input: $bulkEditTarget) {
    id
    targets {
      science {
        name
        tracking {
          __typename
          ... on Sidereal {
            coordinates {
              ra { hms }
              dec { dms }
            }
            parallax { microarcseconds }
          }
        }
      }
    }
  }
}
```
Variable - changes RA to 2 hours and Dec to 2 degrees, removes parallax.

```
{
  "bulkEditTarget": {
    "select": {
      "observationIds": [ "o-2" ]
    },
    "edit": {
      "name": "NGC 5949",
      "ra": { "hours": 2.0 },
      "dec": { "dms": "02:00:00.00" },
      "parallax": null
    }
  }
}
```

rpiaggio:
After taking a look at the new schema, I have to agree with Todd that I would find it a bit more clear to follow the pattern of separating `select` and `edit` objects at all levels.

I think this could be achieved in `updateAllScienceTargets ` by having `EditTargetInput` implement a `select` (`selectTargetName`?) field to specify the target name, and then `nonsiderealEdit`& `siderealEdit` fields, whose `name` fields would be an edit field, just like the rest of the fields. This would remove the need for a `rename` field.

In `updateSiderealScienceTarget`, a `targetName` field can be added to `SelectObservationsInput`.

toddburnside:
While I was writing a longer comment, Raul made a shorter one which is exactly what I was getting at for a solution. What I find really confusing about the current implementation is having a selector field (the name) buried amongst a bunch of fields that will be edited. I don't find it confusing to have the target name in both the `select` and `edit` because it seems logical to me that select happens before edit. Calling it `targetName` in the selector also helps to clarify that.

swalker2m:
@rpiaggio @toddburnside  To make the proposal concrete, you would turn the bulk edit of all science targets into something like this?

```
{
  "bulkEditScienceTargets": {
    "select": {
      "observationIds": [ "o-2" ]
    },
    "edit": {
      "editList": [
        {
          "edit": {
            "select": "Bif",
            # here the fact that it is sidereal is irrelevant
            # but we're trying to avoid a special "rename"
            "siderealEdit": {
              "name": "Bif"
            }
          }
        },
        {
          "add": {
            "sidereal": {
              # since we're adding, no select?
              "name": "Zero",
              "ra": {
                "hms": "00:00:00.00"
              },
             "dec":{
                "dms": "00:00:00.00"
              }
            }
          }
        },
        {
          "edit": {
            "select": "NGC 5949",
            "siderealEdit": {
              "ra": {
                "hms": "11:11:11.11"
              }
            }
          }
        }
      ]
    }
  }
}
```

Or maybe this?  It renames "name" to "select" to leave "name" open for editing.

```
{
  "bulkEditScienceTargets": {
    "select": {
      "observationIds": [ "o-2" ]
    },
    "edit": {
      "editList": [
        {
          "edit": {
            # here the fact that it is sidereal is irrelevant
            # but we're trying to avoid a special rename
            "sidereal": {
              "select": "Bif",
              "name": "Bid"
            }
          }
        },
        {
          "add": {
            "sidereal": {
              # since we're adding, no select?
              "name": "Zero",
              "ra": {
                "hms": "00:00:00.00"
              },
             "dec":{
                "dms": "00:00:00.00"
              }
            }
          }
        },
        {
          "edit": {
            "sidereal": {
              "select": "NGC 5949",
              "ra": {
                "hms": "11:11:11.11"
              }
            }
          }
        }
      ]
    }
  }
}
```

swalker2m:
I suppose if we do one of those, the first `select` could be `selectObservations` and the inner one `selectTarget`.  I think I would make them work all the same instead of having a special single-target case in the API and implementation.

toddburnside:
I think having the select target name outside of the `sidereal` is nicer. That way the `sidereal` is just values being edited. And calling it `selectTargetName` like Raul suggested seems good. Something like:
```
"edit": {
  "selectTargetName": "NGC 5949",
  "sidereal": {
    ...
  }
}

**Comment by toddburnside** [comment]
>  I think I would make them work all the same instead of having a special single-target case in the API and implementation.

Yeah, there is no real reason to have the separate `updateSiderealScienceTarget` other than a small amount of convenience. It is just a subset of `updateAllScienceTargets`.

**Comment by rpiaggio** [comment]
Agree on having the select outside the edit object, and having just one entry point. We didn't mention it but also we wouldn't need `updateScienceTargetName` in this scenario.

**Comment by swalker2m** [comment]
Putting it all together then, we should shoot for this (which I agree seems better):

```json
{
  "singleScienceTarget": {
    "selectObservations": {
      "observationIds": [ "o-2" ]
    },

    "edit": {
      "selectTarget": "NGC 5949",

      "sidereal": {
        "ra": { "hours": 2.0 },
        "dec": { "dms": "02:00:00.00" },
        "parallax": null
      }
    }
  },


  "allScienceTargets": {
    "selectObservations": {
      "observationIds": [ "o-2" ]
    },
    "edit": {
      "editList": [
        {
          "add": {
            "sidereal": {
              "name": "Zero",
              "ra": {
                "hms": "00:00:00.00"
              },
             "dec": {
                "dms": "00:00:00.00"
              }
            }
          }
        },
        {
          "edit": {
            "selectTarget": "Bif",
            "sidereal": {
              "name": "Bid"
            }
          }
        },
        {
          "edit": {
            "selectTarget": "NGC 5949",
            "sidereal": {
              "ra": {
                "hms": "11:11:11.11"
              }
            }
          }
        },
        {
          "delete": {
            "selectTarget": "Foo"
          }
        }
      ]
    }
  }
}
```

Thanks for taking a closer look at this and for all the great suggestions!

toddburnside:
> Putting it all together then, we should shoot for this (which I agree seems better):

This works for me. 😄

swalker2m:
The next question is what to do with "errors".  As implemented if you, say, try to delete a target that doesn't exist in a target list then it will generate an error.  That may sound right, but you may instead want to delete "NGC 5949" in all observations where it is found and not worry about observations where it is missing.  As is you have to first _find_ those observations.  If it were at least an option, you could instead 

```json
{
  "deleteEverywhere": {
    "selectObservations": {
      "programId": "p-2"
    },
    "edit": {
      "editList": [
        { "delete": { "selectTarget": "NGC 5949" } }
      ]
    }
  }
}
```
and not worry about it.

toddburnside:
> The next question is what to do with "errors". As implemented if you, say, try to delete a target that doesn't exist in a target list then it will generate an error. That may sound right, but you may instead want to delete "NGC 5949" in all observations where it is found and not worry about observations where it is missing. As is you have to first _find_ those observations. If it were at least an option, you could instead

From the standpoint of Explore, we will probably only ever call delete for a single observation. But, I can see how not generating an error might be good for API users.

modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
swalker2m: Yes thanks, I believe you're right.  It should be the `NonEmptyStringType` scalar.

swalker2m:
This was a useful discussion and I've implemented the suggested changes, which make the example queries obsolete.  I'm going to close this and open a new clean PR after a few additional changes.

PR #340, opened 2021-08-05
add ws tests for simple queries

tpolecat:
The existing tests run queries over the `http://` endpoint. With this change we also run each query over the `ws://` endpoint. So there's twice as much output now.

 ```
test.QuerySuite:
  • Server starting...
  • Server started.
  + [http] query Programs { ... 1.642s
  + [ws]   query Programs { ... 0.238s
  + [http] query Observations { ... 0.05s
  + [ws]   query Observations { ... 0.051s
  + [http] query ObservationsByConstraintSet { ... 0.06s
  + [ws]   query ObservationsByConstraintSet { ... 0.048s
  + [http] query ObservationsByScienceRequirements { ... 0.046s
  + [ws]   query ObservationsByScienceRequirements { ... 0.042s
  • Server stopped.
Execution took 2.18s
8 tests, 8 passed
All tests in test.QuerySuite passed
```

This is really just a sanity check for the routing logic and it might make sense to remove it at some point, but for now I think it's useful to check that they're consistent.


swalker2m (approved):
Looks great!  :+1:

PR #341, opened 2021-08-06
Allow expecting error messages

swalker2m:
Adds an update to `OdbSuite` to allow testing queries that are expected to fail.  It also introduces some cruft that will be removed when the next clue update is published.


cquiroz (approved):
LGTM

cquiroz (approved):
LGTM

PR #353, opened 2021-08-12
No sharing targets

swalker2m:
Following up on the initial offer in #338, this PR implements a few recommended changes to the no-sharing API.  Again, the "pointing" (either asterism or target) has been replaced with a `TargetEnvironmentModel` that contains an explicit base position (optional) and a list of science targets.  It would eventually house guide stars if we need to store them and blind offset stars, etc.

The PR is impossibly big but there are a series of example queries and mutations at the end which may be of use.  See

* `TargetQuerySuite`
* `TargetQuerySuite`
* `ScienceTargetMutationSuite`
* `ScienceTargetsMutationSuite`
* `TargetEnvironmentMutationSuite`

For example, a science target edit that happens to be a rename of target "NGC 3312":

```
  // Rename target "NGC 3312" to "NGC 3312*" in o-3.
  queryTest(
    query ="""
      mutation UpdateScienceTarget($renameEdit: BulkEditScienceTargetInput!) {
        updateScienceTarget(input: $renameEdit) {
          id
          targets {
            science {
              name
            }
          }
        }
      }
    """,
    expected =json"""
      {
        "updateScienceTarget": [
          {
            "id": "o-3",
            "targets": {
              "science": [
                {
                  "name": "NGC 3312*"
                }
              ]
            }
          }
        ]
      }
    """,
    variables =json"""
      {
        "renameEdit": {
          "selectObservations": [ "o-3" ],
          "edit": {
            "selectTarget": "NGC 3312",
            "sidereal": {
              "name": "NGC 3312*"
            }
          }
        }
      }
    """.some,
    clients = List(ClientOption.Http)  // cannot run this test twice since it changes required state
  )
```


cquiroz (approved):
I confess I didn't read it all but concentrated on the examples and tests and it seems good

toddburnside:
I took the same path as Carlos. The examples and tests look good. If we leave this as a PR, I can create a review app and work on the Explore side.

PR #356, opened 2021-08-13
add a subscription test

tpolecat:
After being stumped for a while because of a dumb mistake, here is a basic subscription test. The ergonomics may not be good in the long run but it's a start. I left in some debug logging because I suspect we'll need it.


swalker2m (approved):
This looks great!  Assuming the target API changes (#353) this will have to be switched to an observation edit or something, but I can do that.  :+1:

PR #357, opened 2021-08-16
Switch target subscription to observation subscription

swalker2m:
Switches the target subscription test to an observation subscription test in order to facilitate [pending updates in the target API](https://github.com/gemini-hlsw/lucuma-tmp-api/pull/353).


cquiroz (approved):
LGTM

PR #366, opened 2021-08-30
Constraint Set Evolution

swalker2m:
Removes the constraint set name, adds a `Default` constraint set model that matches the UI and switches the observation model default to use it.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
@@ -7,10 +7,8 @@ import cats._
 import cats.data.State
 import cats.syntax.all._
 import clue.data.Input
-import eu.timepit.refined.types.string.NonEmptyString
 import io.circe.Decoder
 import io.circe.generic.semiauto._
-import io.circe.refined._
 import lucuma.core.enum._
 import lucuma.core.optics.syntax.lens._
 import lucuma.odb.api.model.syntax.input._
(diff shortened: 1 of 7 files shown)


cquiroz (approved):
LGTM. Is the generated name built on the client?

swalker2m:
> Is the generated name built on the client?

It must be at the moment.  I could add it to the schema though if that would be useful.

toddburnside:
Yes. The name is built on the client. The only reason I can see adding it to the schema would be if API users wanted/needed it. Which, I don't think is necessary.

modules/core/src/main/scala/lucuma/odb/api/model/ConstraintSetModel.scala
```diff
@@ -31,18 +28,31 @@ object ConstraintSetModel extends ConstraintSetModelOptics {
...
+  /**
+   * Constraints that are set by default when a new observation is created.
+   */
+  val Default: ConstraintSetModel =
+    ConstraintSetModel(
+      ImageQuality.PointEight,
+      CloudExtinction.PointThree,
+      SkyBackground.Bright,
+      WaterVapor.Wet,
+      ElevationRangeModel.airmassRange.reverseGet(AirmassRange.AnyAirmass)
```
toddburnside, line 52: It looks like this is Airmass 1.0 - **3.0**. I think Andy wants 1.0 - **2.0**

swalker2m: Thanks, that makes more sense for a default.  I'll put in a new PR for it.

PR #377, opened 2021-09-09
Core update

cquiroz:
There is an experimental export of the `basic-case` api that needs to be reworked and it is unused at the moment. I'm removing it for a while and to make it easier to update lucuma-core


swalker2m (approved):
Looks good to me :+1:

PR #390, opened 2021-10-01
Unaffiliated Target Environments, etc.

swalker2m:
I've reached a hopefully somewhat complete stopping point in a big refactoring to support unaffiliated targets and target environments.  It wants polishing but in theory could be used to update the UI.  

* The PR is artificially big because I separated all the classes that previously were inside `object TargetModel`.
* Target query results are not paged.  I will either add paging here or else remove it elsewhere.
* Observation editor doesn't directly include a target environment editor.  Instead you have dedicated target editors.

Because the PR is unworkably big, it's probably easiest to look at the test suites at the bottom to see the queries and results.


swalker2m:
A couple more notes:

* At the moment, the client can suggest an ID for target environments but not for targets.  The issue with targets is that you can specify a target to be created in multiple environments at once.  Copies of the target are made, each with its own ID.
* There's a target environment create/edit event but i haven't made a test suite for it yet.

PR #410, opened 2021-10-29
Add error to magnitudes for queries

toddburnside:
It was possible to specify a Magnitude error when creating a Magnitude, but they were not available in queries.


cquiroz (approved):
Cool

PR #418, opened 2021-11-11
Avoid a needless warning

swalker2m:
These lines produced some superfluous (?) warnings:

```
[warn] /Users/swalker/dev/lucuma-tmp-api/modules/service/src/test/scala/OdbSuite.scala:45:11: private method require in <$anon: lucuma.sso.client.SsoClient[cats.effect.IO,lucuma.core.model.User]> is never used
[warn]       def require(req: Request[IO])(f: User => IO[Response[IO]]): IO[Response[IO]] = locally(???)
[warn]           ^
```

This PR turns them off but I don't understand why there were warnings about unused methods that are required in order to implement an interface.   What am I doing wrong?


swalker2m:
Hmm and now it works locally but not on GitHub.

PR #422, opened 2021-11-11
Update to lucuma-core 0.14.3

swalker2m:
Minimal changes required to update to `lucuma-core` 0.14.3 and its ADT `Target` model.  My plan is to make corresponding changes to the GraphQL schema in a subsequent PR.


modules/core/src/main/scala/lucuma/odb/api/model/targetModel/TargetModel.scala
```diff
@@ -41,46 +37,10 @@ trait TargetModelOptics { self: TargetModel.type =>
...
-  val ra: Optional[Target, RightAscension] =
-    coordinates.andThen(Coordinates.rightAscension)
-
-  val dec: Optional[Target, Declination] =
-    coordinates.andThen(Coordinates.declination)
-
-  val epoch: Optional[Target, Epoch] =
-    siderealTracking.andThen(SiderealTracking.epoch)
+  // Target.properMotion is Optional[Target, ProperMotion], which makes it
+  // impossible to assign None to the ProperMotion of a SiderealTarget ?
```
swalker2m, line 41: Am I missing something here?

PR #449, opened 2021-12-14
Shared Targets

swalker2m:
First big step refactoring for unshared asterisms / shared targets.  It's already un-reviewable unfortunately but the API updates can be seen in the `test.targets` cases.  Remaining steps include:

* [ ] Additional validation on target id assignment
* [x]  By-target grouping in the schema (asterism and environment grouping are present)
* [ ] Easier asterism editing (for the moment you have to assign the entire target list)
* [ ] Support for cloning targets
* [ ] `observed` attribute / validation that observed targets cannot be edited


toddburnside (commented):
You're right - it is unreviewable. :) I looked it over and didn't see any glaring issues.  I think we should go the route of the last major rewrite - create a review app and start in on explore. It'll be easier to assess once I have the playground and start writing the queries.

swalker2m:
@toddburnside I added `groupByTarget` which required some changes because it's the first case of grouping by something which potentially is not associated with any observations.

It won't win any awards for cleverness or efficiency or anything else but it'll all be replaced by the definitive database-backed version soon anyway.  If you don't see any glaring issues maybe we should just merge it and then subsequent steps can be more manageable PRs?

toddburnside:
> @toddburnside I added `groupByTarget` which required some changes because it's the first case of grouping by something which potentially is not associated with any observations.
> 
> It won't win any awards for cleverness or efficiency or anything else but it'll all be replaced by the definitive database-backed version soon anyway. If you don't see any glaring issues maybe we should just merge it and then subsequent steps can be more manageable PRs?

I guess my question is whether we want to merge and use the development server, or use a review app. The former would block any unrelated changes from being merged (at least to staging) until I get Explore working with the changes. The latter might result in more merge conflicts for you.

Hopefully I'll be able to have a minimally working explore by early next week (worst case I comment out the target tab, etc.). So maybe merging is OK.

toddburnside:
> Hopefully I'll be able to have a minimally working explore by early next week (worst case I comment out the target tab, etc.). So maybe merging is OK.

I think Raul will be embarking on changes for target brightness in the API soon, so maybe we should merge and deal with a potential delay in Explore. The merge conflicts could be bad.

toddburnside (approved):
LGTM

swalker2m:
> I think Raul will be embarking on changes for target brightness in the API soon, so maybe we should merge and deal with a potential delay in Explore. The merge conflicts could be bad.

I too was worried about that conflict.  Raul and I discussed it and I'll be adding the brightness changes on top of this.

PR #452, opened 2021-12-16
Validate assigned target's program

swalker2m:
Checks that all targets assigned to an observation reference the same program.


toddburnside (approved):
LGTM

PR #456, opened 2021-12-17
Clone target in

swalker2m:
With the goal of making it possible to clone a target in given list of observations, this PR:

* Adds an `updateAsterism` mutation.  You can list multiple fine grained edits to perform in a single step.  Before the only option was to replace the entire asterism list with another.  E.g. swapping target `t-2` in for `t-4`:

```
      mutation ReplaceTargets($listEdit: BulkEditAsterismInput!) {
        updateAsterism(input: $listEdit) {
          id
          targets {
            asterism {
              name
            }
          }
        }
      }

      {
        "listEdit": {
          "selectObservations": [ "o-3", "o-4" ],
          "edit": [
            {
              "delete": "t-4"
            },
            {
              "add": "t-2"
            }
          ]
        }
      }
```

* Fixes a bug in which the cloned target wasn't being stored

* Adds an optional `observationIds` parameter to `cloneTarget`.  If specified it can list observations for which the cloned target should be swapped in for the existing target that was cloned.  For example this mutation would replace `t-4` in observations `o-3` and `o-4` with the new clone `t-a`.  The `suggestedCloneId` can be left out to assign a target id automatically.  `observationIds` can be left out to only perform the clone and not update any existing observations.

```
      mutation CloneAndReplaceTarget {
        cloneTarget(existingTargetId: "t-4", suggestedCloneId: "t-a", observationIds: [ "o-3", "o-4" ]) {
          id
          name
        }
      }
```


toddburnside (approved):
Nice. These new features will be very helpful in the UI.

PR #465, opened 2021-12-29
SourceProfile output types

swalker2m:
This is the first in a series of updates required to add `SourceProfile` to the schema.  It includes only the output types for now.  Future PRs would allow you to create and edit the source profile but at this point you can query any existing profile information.  Adding `SourceProfileSchema` is the goal of this PR, but it also:

1. tracks a more recent `lucuma-core`.  As such it has to invent a `SourceProfile` for new targets created without any such information.  Because we require a `SourceProfile` this will be updated in the future to require the corresponding input when a target is created.  For now, it makes a bogus `Point` source.
2. turns off SSO again, for now.  It pulls in obsolete dependencies.

## Ponderous Queries 

The queries are somewhat ponderous, being a straightforward mapping to the `SourceProfile` model.  Maybe that is not ideal?

```
    ... on PointSource {
      spectralDefinition {
        ... on BandNormalizedIntegrated {
          sed {
            ... on Planet {
              spectrum
            }
          }
        }
      }
    }
```

## Inconsistent units handling

It pains me that units are different for magnitudes, line flux, etc.  Ordinarily when units are involved the query allows the caller to get the value in the units they choose:

Screen Shot 2021-12-29 at 11 11 49

Here you can select `arcsec` or `degrees` or whatever and see the corresponding value.  For the magnitudes I'm not sure whether the options are really even convertible in all cases so you just get a (value, units) pair essentially:

Screen Shot 2021-12-29 at 11 11 02


modules/core/src/main/scala/lucuma/odb/api/schema/SourceProfileSchema.scala
```diff
@@ -0,0 +1,627 @@
...
+          resolve     = _.value.fluxDensities.toNel.toList
+        )
+      )
+    )
+
+  // We are limited in the characters we can use for enum names.  These mappings
+  // define how to map illegal characters into something valid for GraphQL.
+  private val replacements: List[(String, String)] =
+    List(
+      " " -> "_",
```
cquiroz, line 233: Nice

cquiroz (approved):
LGTM

PR #466, opened 2021-12-29
Sample profiles

swalker2m:
Hacks in a way to set the `SourceProfile` in the sample targets in `Init`.  I don't know whether they correspond to reality but it's good to have something set for those targets.

The source profile information will be added to the GraphQL input and the hack will be removed in a subsequent PR.


cquiroz (approved):
LGTM

PR #467, opened 2021-12-29
Report validation errors

swalker2m:
A small change that would have prevented some grief.  By moving the schema construction outside the function `WebSocketBuilder2 => HttpApp` a validation failure occurs in the `Resource` map and gets reported instead of silently eaten.


toddburnside (approved):
LGTM

PR #475, opened 2022-01-04
Tiny leftover

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.quiroz@noirlab.edu>


swalker2m (approved):
:+1: I think I did that on purpose for the documentation thinking the term "magnitude" bands is more commonly employed.  Maybe better to be consistent though.

PR #476, opened 2022-01-04
lucuma-core 0.22.0 and some source profile inputs

swalker2m:
Lazily combines a couple of unrelated things:

* Starting to define some source profile inputs (for creation and editing)
* Updates to `lucuma-core` 0.22.0 and makes corresponding changes

The first item is still very much WIP and will continue to evolve.


rpiaggio (approved):
LGTM

PR #480, opened 2022-01-06
Source profile inputs 2

swalker2m:
Inputs for creating source profile.  This is a lot of boring rote code but you can get a glimpse of the end result in the `Init` and `TestInit` code that sets up a test "database".


modules/service/src/main/scala/lucuma/odb/api/service/Init.scala
```diff
@@ -42,7 +38,89 @@ object Init {
...
+          "galaxy": "SPIRAL"
+        },
+        "brightnesses": [
+          {
+            "brightness": {
+              "value": 12.7,
+              "units": "VEGA_MAGNITUDE"
+            },
+            "band": "B"
+          },
```
swalker2m, line 55: Here a point source is created, for example.

modules/service/src/main/scala/lucuma/odb/api/service/Init.scala
```diff
@@ -42,7 +38,89 @@ object Init {
...
+          {
+            "brightness": {
+              "value": 10.279,
+              "units": "VEGA_MAGNITUDE"
```
swalker2m, line 59: On the input (i.e., here) I could turn this around

```
"brightnesses": [
  {
    "vega": 10.279,
    "band": "B"
  },
  ...
]
```

 to match the way it is done elsewhere (like for "parallax" above):

```
"parallax": { 
  "milliarcseconds": 1.0
}
```

I like that better and it is more consistent.  The problem is that on output (that is, when queried) I cannot support getting the value in any other units except those in which it was entered. I don't think we can easily translate between them.   Whereas for parallax we can cleverly provide the value in μas even though it was entered as mas:

```
"parallax": {
  "microarcseconds": 1000
}
```
I don't think we can do that for brightness. You couldn't take that `10.279` value and get the equivalent (if there is an equivalent) in, say, `erg/s/cm²/Hz`.   So the output would, inconsistently, continue to be "value" and "units".    :-/

cquiroz: Right, some of the units can be converted but not every one. Also you can't convert surface to integrated or viceversa.

cquiroz (approved):
LGTM

PR #486, opened 2022-01-10
No union source profile

swalker2m:
I'm not sure where this falls on the spectrum from dumb to good so I'd like some input.  The idea is to make the source profile query output match the JSON input required to create it.  Eventually this idea would be applied to the entire `Target` type.

So practically speaking this means removing all the GraphQL `union`s that would otherwise be present.  Instead all the possible values turn into fields with `Option` values.  A few tradeoffs occur to me:

### Advantages

* If what you get from a query matches the form and content of what you need to do to create an object, then it makes copy/paste/edit script or playground sessions easier?  
* You don't need to see the input type if you know the output type and vice-versa.
* Less verbose / cumbersome than the `... on Whatever` query syntax

### Disadvantages

* You can not tell that exactly one of the cases will always be present.  `SourceProfile` for example would need option `point`, `uniform` and `gaussian` fields and exactly one of them would be present, though there's no way to express that in the SDL.  
* The `__typename` field won't help with parsing query results.

### Example

Ignoring the brightnesses for conciseness, instead of

```
sourceProfile {
  ... on PointSource {
    spectralDefinition {
      ... on BandNormalizedIntegrated {
        sed {
          ... on Galaxy {
            spectrum
          }
        }
      }
    }
  }
}
```

you could write 

```
sourceProfile {
  point {
     bandNormalized {
       sed {
         galaxy
       }
     }
  }
}
```

and the output looks like the values you need for input 

```
"sourceProfile": {  
  "point": {
    "bandNormalized": {
      "sed": {
        "galaxy": "ELLIPTICAL"
      }
    }
  }
}
```

Of course for a random target it may not be `point`/`bandNormalized`/`galaxy` in which case you get `null` values at the point where it doesn't match.  For example, if this were a uniform source you'd see

```
"sourceProfile": {
  "point": null
}
```

This doesn't seem any worse than when `union`s are used.  If you want to be prepared for anything you have to expand all the cases either way.

So, is this dumb?

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/schema/SourceProfileSchema.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/schema/SourceProfileSchema.scala
@@ -11,15 +11,14 @@ import lucuma.core.math.{BrightnessValue, Wavelength}
 import lucuma.core.math.BrightnessUnits.{Brightness, FluxDensityContinuum, Integrated, LineFlux, Surface}
 import lucuma.core.math.dimensional.{Measure, Of, Units}
 import lucuma.core.model.SpectralDefinition.{BandNormalized, EmissionLines}
-import lucuma.core.model.UnnormalizedSED.{BlackBody, CoolStarModel, Galaxy, HIIRegion, Planet, PlanetaryNebula, PowerLaw, Quasar, StellarLibrary, UserDefined}
 import lucuma.core.model.{SourceProfile, SpectralDefinition, UnnormalizedSED}
 import lucuma.core.util.Enumerated
 import lucuma.odb.api.model.targetModel.SourceProfileModel.{BandBrightnessPair, CreateBandBrightnessInput, CreateBandNormalizedInput, CreateEmissionLineInput, CreateEmissionLinesInput, CreateGaussianInput, CreateMeasureInput, CreateSourceProfileInput, CreateSpectralDefinitionInput, FluxDensityInput, UnnormalizedSedInput, WavelengthEmissionLinePair}
+import monocle.Prism
 import sangria.schema.{Field, _}
 import sangria.macros.derive._
 import sangria.marshalling.circe._
 
-import scala.reflect.ClassTag
 
 object SourceProfileSchema {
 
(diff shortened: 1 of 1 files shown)


swalker2m:
I think you can get the output-matches-input feature without losing the benefits of `union`s, though it is still more verbose.

```
sourceProfile {
  ... on PointSource {
    point {
      ... on BandNormalizedIntegrated {
        sed {
          ... on Galaxy {
            galaxy
          }
        }
      }
    }
  }
}
```

swalker2m:
I think you can get the input-matches-output feature without losing the benefits of `union`s but it requires some reworking of the types in an awkward and even more verbose way?  Instead of a `PointSource` with a `spectralDistribution` field you'd rename the field to `point`.  Similarly the `Galaxy` type would need a `galaxy` field which has the enum value.

`BandNormalizedIntegrated` combines the SED with the magnitudes and would need to be a type that _has_ a `bandNormalized` field which in turn has the `sed` and `brightnesses`.

```
sourceProfile {
  ... on PointSource {
    point {
      ... on BandNormalizedIntegrated {
        bandNormalized {
          sed {
            ... on Galaxy {
              galaxy
            }
          }
        }
      }
    }
  }
}
```

rpiaggio:
At first glance, this looks like a great improvement. It makes queries more readable and manageable.

The resulting schema is looser, allowing representations of impossible states; but the actual onus of honoring the "oneOf" contract falls on the server.

Clients are not affected that much: as Tood mentioned we use "trial and error" to decode sum types. We would just have to add one line to the specific type decoders, eg: `c.downField("point")` and it should work. There's an extra advantage: we can use the repeated field names in the different sum type instances (eg: `galaxy.spectrum` and `planet.spectrum`), which we couldn't do with union types.

Maybe, for schema documentation purposes, we can add a `@oneOf` schema directive/annotation to these types?

swalker2m:
> Maybe, for schema documentation purposes, we can add a `@oneOf` schema directive/annotation to these types? (we probably want that in the input types too).

I think that's a good idea.  I played with it a bit in Sangria without much success. It may be something to consider when we switch to Grackle. 

Since I think the option value approach is somewhat better than the `union`s, at least for `SourceProfile`, I'll go ahead and merge this unless there are any objections?

rpiaggio:
Ok for me.

toddburnside (approved):
OK by me, too.

PR #490, opened 2022-01-13
Target model consistency

swalker2m:
Some target schema cleanup in preparation for editing source profiles.  It

* makes the create input and the output of a target query similar by removing the sidereal/nonsidereal tracking union in favor of `sidereal` / `nonsidereal` optional fields on `Target`
* adds catalog info object type and edit support and moves catalog info into sidereal to match the underlying model
* adds full angular size support

Catalog info and angular size, in addition to being parts of the target model that needed support, offered a mini-case of kinds of thing that will need to be done in the source profile editor.  Namely, I needed information about the existing thing being edited in order to validate arguments.  For example, if you don't supply both a major and minor axis for the angular size, you can't create a new angular size.  If you're just editing an existing angular size though you can update just major or minor and not touch the other.  Knowing how to validate the arguments depends on knowing whether you are trying to create a new angular size or just editing an existing one.

So instead of `ValidatedInput[State[A, Unit]]` for an editor of `A` it becomes a `StateT[EitherNec[InputError, *], A, Unit]`.  I'm planning on propagating that change outside of the target model but for now it's only done there. This required adding special `Lens`/`Optional` state syntax.  Eventually I'll no longer need the version in `lucuma-core`.


modules/core/src/main/scala/lucuma/odb/api/schema/TargetSchema.scala
```diff
@@ -285,65 +306,70 @@ object TargetSchema extends TargetScalars {
...
+          description = "Right ascension at epoch".some,
+          resolve     = _.value.tracking.baseCoordinates.ra
+        ),
+
+        Field(
+          name        = "dec",
+          fieldType   = DeclinationType[F],
+          description = "Declination at epoch".some,
+          resolve     = _.value.tracking.baseCoordinates.dec
         ),
```
swalker2m, line 333: Replaced `coordinates` with `ra` and `dec` here to match the input.

modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
```diff
@@ -225,8 +236,8 @@ trait TargetMutation extends TargetScalars {
       name        = "createTarget",
       fieldType   = TargetType[F],
       description = "Creates a new target according to the provided parameters.  Only one of sidereal or nonsidereal may be specified.".some,
-      arguments   = List(ArgumentTargetCreate),
-      resolve     = c => c.target(_.insert(c.arg(ArgumentTargetCreate)))
+      arguments   = List(ProgramIdArgument, ArgumentTargetCreate),
```
swalker2m, line 239: Took the `programId` out of the target creation arguments.  You should be able to use the same parameters to create targets in multiple programs, I thought.

modules/core/src/main/scala/lucuma/odb/api/model/targetModel/TargetModel.scala
```diff
@@ -121,29 +131,52 @@ object TargetModel extends TargetModelOptics {
...
+          // Create or edit, depending on whether it already exists
+          case Assign(editor) =>
+            StateT.modifyF[EitherNec[InputError, *], Target] { t =>
+              Target
+                .angularSize
+                .modifyF[EitherNec[InputError, *]](
+                  _.fold(editor.create.toEither)(as => editor.edit.runS(as)).map(_.some)
+                )(t)
+            }
+        }
```
swalker2m, line 168: This pattern of handling `Input` repeats.  I will try to clean it up and abstract it.

modules/core/src/main/scala/lucuma/odb/api/model/syntax/LensOps.scala
```diff
@@ -0,0 +1,60 @@
...
+    toState
+
+  def extract: StateT[EitherInput, S, A] =
+    toState
+
+  def edit(a: A): StateT[EitherInput, S, A] =
+    assign(a)
+
+  @inline def :=(a: A): StateT[EitherInput, S, A] =
+    edit(a)
```
swalker2m, line 25: It seemed better to tie down the effect in `StateT` than to add it to the `:=` method and ruin the syntax.  I don't know.  I guess it would get inferred.

modules/core/src/main/scala/lucuma/odb/api/model/NullableInput.scala
```diff
@@ -0,0 +1,58 @@
...
+
+      // Create a new instance of A for the state S or else edit the existing
+      // A instance in the state S.  Arguments to the input are validated
+      // appropriately either way.
+      case Assign(n) =>
+        StateT.modifyF[EitherInput, S] { s =>
+          optional
+            .modifyA[EitherInput](_.fold(n.create.toEither)(n.edit.runS).map(_.some))(s)
+        }
+    }
```
swalker2m, line 56: I'm not proud of `NullableInput` or the way this is written but a better way hasn't occurred to me yet. Advice appreciated.  At any rate, it does accomplish the goal of not repeating essentially the same code everywhere it is needed (for angular size and catalog info so far).

rpiaggio:
I haven't looked at the new editing logic yet, but seeing the angular size here makes me wonder if it belongs in the model at all. 

Its only use is for the UI to default to a nice visualization angle when performing catalog queries. It is never edited nor queried afterwards. As such, it was never meant to be persisted and it probably belongs to a new `CatalogResult(target, angularSize)` in `lucuma-catalogs` instead of the target model.

swalker2m:
> it was never meant to be persisted and it probably belongs to a new CatalogResult(target, angularSize) in lucuma-catalogs instead of the target model

Ah ha, yes I think that makes more sense.  I guess I won't remove it from the API until it is gone from the model but that makes sense.

PR #492, opened 2022-01-17
Use StateT across ObservationModel

swalker2m:
This is an implementation change, but not a change to the input or output GraphQL types.  It uses `StateT` editors across the `ObservationModel` hierarchy for consistency and flexibility.  At the "repo" level there is still a mix of `ValidatedInput[State ...]]` and `StateT` which is a bit awkward, but  I plan to address that.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/syntax/PrismOps.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/syntax/PrismOps.scala
@@ -22,14 +22,6 @@ final class PrismOps[S, A](val self: Prism[S, A]) extends AnyVal {
       }
     }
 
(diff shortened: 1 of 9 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/syntax/PrismOps.scala
```diff
@@ -0,0 +1,48 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.odb.api.model.syntax
+
+import cats.syntax.either._
+import cats.data.StateT
+import lucuma.odb.api.model.{EitherInput, InputError}
+import monocle.Prism
+
+final class PrismOps[S, A](val self: Prism[S, A]) extends AnyVal {
```
cquiroz, line 11: Nice

PR #494, opened 2022-01-18
Nonsidereal key

swalker2m:
A small-ish schema change to output/accept a `key` field for ephemeris when showing/creating a nonsidereal target.  Combines the sidereal create/edit inputs into one input type (and does the same for nonsidereal).  Also, allows switching from sidereal to nonsidereal, fwiw.


cquiroz (approved):
LGTM

PR #496, opened 2022-01-19
Editable source profile (Part 1)

swalker2m:
Swaps `CreateSourceProfileInput` for `SourceProfileInput` which will be used for both creating and editing.  Adds a `sourceProfile` field to the target editor.  At the moment, you have to replace the entire source profile if you want to replace any part of it.  That will be addressed in the next PR, when it comes.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/syntax/LensOps.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/syntax/LensOps.scala
@@ -12,6 +12,9 @@ final class LensOps[S, A](val self: Lens[S, A]) extends AnyVal {
   def transform(st: StateT[EitherInput, A, Unit]): StateT[EitherInput, S, Unit] =
     st.transformS[S](self.get, (s, a) => self.replace(a)(s))
 
+  def :<(st: Option[StateT[EitherInput, A, Unit]]): StateT[EitherInput, S, Unit] =
+    st.fold(StateT.empty[EitherInput, S, Unit])(transform)
+
   def toState: StateT[EitherInput, S, A] =
     StateT(s => Right((s, self.get(s))))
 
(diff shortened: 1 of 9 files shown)


modules/core/src/main/scala/lucuma/odb/api/schema/TargetMutation.scala
```diff
@@ -149,22 +149,26 @@ trait TargetMutation extends TargetScalars {
 
   implicit val InputObjectEditTargetInput: InputObjectType[TargetModel.Edit] = {
 
-      // Not able to derive this for some reason, TBD.
+    // Not able to derive this for some reason, TBD.
 //    deriveInputObjectType[TargetModel.Edit](
 //      InputObjectTypeName("EditTargetInput"),
```
swalker2m, line 154: I haven't figured this out yet.  The error message is `not found: value Edit`, though the same construct works everywhere else and I can't spot any difference.  I assume it has something to do with macros because `TargetModel.Edit` is visible just below in the non-macro version.

modules/core/src/main/scala/lucuma/odb/api/model/syntax/LensOps.scala
```diff
@@ -12,6 +12,9 @@ final class LensOps[S, A](val self: Lens[S, A]) extends AnyVal {
   def transform(st: StateT[EitherInput, A, Unit]): StateT[EitherInput, S, Unit] =
     st.transformS[S](self.get, (s, a) => self.replace(a)(s))
 
+  def :<(st: Option[StateT[EitherInput, A, Unit]]): StateT[EitherInput, S, Unit] =
```
cquiroz, line 15: Nice operator

cquiroz (approved):
LGTM, @rpiaggio I guess this triggers a schemas update?

swalker2m:
> I guess this triggers a schemas update?

There's more to come, unfortunately, so you may want to wait.

PR #498, opened 2022-01-20
Editable source profile (Part 2)

swalker2m:
Next installment in editable source profiles.  Continues turning the `CreateFooInput`s that work for creation only into `FooInput`s that work for creation or editing.  All the fields becomes optional but on creation required fields are checked and flagged if missing.  On editing you only need to specify the things that are changing.  For example, switching from one type of `SourceProfile` to another (e.g., `Point` to `Gaussian`) requires specifying both the `fwhm` and the spectral definition whereas editing the fwhm of an existing `Gaussian` doesn't require the spectral definition.

I'm working my way down from the top and have `SourceProfileInput`, `GaussianInput`, `SpectralDefinitionInput`, `EmissionLinesInput`, and `BandNormalizedInput` converted so far.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/EditorInput.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/EditorInput.scala
@@ -5,6 +5,7 @@ package lucuma.odb.api.model
 
 import cats.data.StateT
 import cats.syntax.either._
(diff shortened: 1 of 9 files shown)


cquiroz (approved):
LGTM

PR #500, opened 2022-01-21
Editable source profile (Part 3)

swalker2m:
No sense deriving input objects where every field must be replaced.


cquiroz (approved):
LGTM

PR #501, opened 2022-01-21
Editable source profile (Part 4)

swalker2m:
Makes it possible to edit band brightness values.  You reference the bands you want to edit and then supply just the changes.  

For example, to change the value of the "H" magnitude, you don't need to specify the units (they remain unchanged).  Nor do the unchanged magnitudes of other bands need to be mentioned:

```
    "sourceProfile": {
      "point": {
        "bandNormalized": {
          "brightnesses": [
            {
              "band": "H",
              "value": 9.999
            }
          ]
        }
      }
    }
```

To remove just the error value:

```
    "sourceProfile": {
      "point": {
        "bandNormalized": {
          "brightnesses": [
            {
              "band": "H",
              "error": null
            }
          ]
        }
      }
    }
```

of course you can do both at once and edit other magnitudes at the same time

```
    "sourceProfile": {
      "point": {
        "bandNormalized": {
          "brightnesses": [
            {
              "band": "H",
              "value": 9.999
              "error": null
            },
            {
              "band": "K",
              "value": 9.0
            }
          ]
        }
      }
    }
```

When adding a new magnitude, you have to specify all the required fields (i.e., all but `error`):

    "sourceProfile": {
      "point": {
        "bandNormalized": {
          "brightnesses": [
            {
              "band": "U",
              "value": 10.0,
              "units": "VEGA_MAGNITUDE"
            }
          ]
        }
      }
    }
```

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
@@ -31,6 +31,7 @@ import lucuma.core.model.{EmissionLine, SourceProfile, SpectralDefinition, Unnor
 import lucuma.odb.api.model.{AngleModel, EditorInput, EitherInput, InputError, ValidatedInput, WavelengthModel}
 import lucuma.odb.api.model.syntax.input._
 import lucuma.odb.api.model.syntax.lens._
+import monocle.{Focus, Lens}
 
 import scala.collection.immutable.SortedMap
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by swalker2m** [comment]
You cannot delete existing magnitudes altogether.  I think I'll add a field to `bandNormalized` listing the bands you'd like to remove to do that.

**Comment by cquiroz** [comment]
are we actually deleting bands or just marking them as deleted?

**Comment by swalker2m** [comment]
> are we actually deleting bands or just marking them as deleted?

Hmm, at the moment only top-level things with a `Gid` have a deletion flag.  Anything else that can be deleted is really deleted.

**Comment by cquiroz** [comment]
Ok, sounds like a good policy

File: modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala [hunk 1]
```diff
@@ -233,47 +234,77 @@ object SourceProfileModel {
         a.measure
       )}
 
+    def measure[T]: Lens[BandBrightnessPair[T], Measure[BrightnessValue] Of Brightness[T]] =
```
cquiroz, line 237: Should this be in core?

cquiroz (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
swalker2m: I resurrected the `BandBrightnessPair` which was in an earlier incarnation of the source profile model because it was useful to me.  Since the class itself only exists in the API now, the `Lens` has to go there too.

PR #502, opened 2022-01-24
Editable source profile 5

swalker2m:
This installment updates the emission line input to make it editable and uses the editor in its containing emission line*s* input.  This allows adding new emission lines or editing existing ones identifying them by wavelength.  Deletion support, as for bands in band normalized, is pending.


cquiroz (approved):
LGTM

PR #505, opened 2022-01-25
Editable source profile 7

swalker2m:
Proposes adding `editBrightnesses` and `deleteBrightnesses` in addition to `brightnesses` in `BandNormalizedInput`.  So you use `brightnesses` to set the magnitudes to a given list of values.  This matches the `brightnesses` field in the output type.  If you want to make changes to existing brightnesses (adding or modifying) you can instead (or in addition) set `editBrightnesses` to the changes you want to make.  Finally `deleteBrightnesses` can be set the bands of the corresponding entries to remove.

Adds the same pattern for wavelengths and emission lines.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
@@ -31,6 +31,8 @@ import lucuma.core.model.{EmissionLine, SourceProfile, SpectralDefinition, Unnor
 import lucuma.odb.api.model.{AngleModel, EditorInput, EitherInput, InputError, ValidatedInput, WavelengthModel}
 import lucuma.odb.api.model.syntax.input._
 import lucuma.odb.api.model.syntax.lens._
+import lucuma.odb.api.model.targetModel.SourceProfileModel.BandNormalizedInput.brightnessEditor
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
```diff
@@ -296,45 +296,53 @@ object SourceProfileModel {
...
+                 brightnessNoop[T],
+                 StateT.setF(emptyBrightnessMap[T].rightNec[InputError]),
+                 in => StateT.setF(brightnessCreator[T](in).toEither)
+               ).some
+
+        // `editBrightnesses` to edit existing ones or add new ones
+        _ <- BandNormalized.brightnesses[T] :<
+               editBrightnesses.toOption.fold(brightnessNoop[T])(brightnessEditor[T]).some
+
+        // `deleteBrightnesses` to delete existing ones
```
cquiroz, line 339: Maybe we need a `deleteAll`?

cquiroz (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
swalker2m: That might be easier.  I thought about it but you can also accomplish that with setting it to an empty list via `"brightnesses": []`

PR #506, opened 2022-01-27
Constraint set makeover

swalker2m:
Updates the constraint set graph to be consistent with source profile.  Namely:

* Combines the create and edit inputs into a single input that works for either, checking arguments appropriately for each case
* Allows all parts of the tree to be edited independently.  You can update just the `min` airmass range for example.
* Makes the constraint set output shape match the input.  In other words, removes the `Union` in favor of optional `airmassRange` and `hourAngleRange` under `elevationRange`


cquiroz (approved):
Looks great. I guess we need to update lucuma-schemas soon

swalker2m:
> I guess we need to update lucuma-schemas soon

I'm going to make similar changes elsewhere so if you want to wait you might avoid having to do it several times.

PR #507, opened 2022-01-28
liftState

swalker2m:
A minor cosmetic change to take a `ValidatedInput` into a `StateT`.  I don't use it on the individual arguments though, preferring a `(arg1.validate, arg2.validate).tupled.liftState` pattern so that both arguments get checked instead of stopping on the first failure.


cquiroz (approved):
LGTM

PR #512, opened 2022-01-31
Science configuration

swalker2m:
Converts the `ScienceConfiguration` to the same editor pattern used by `SourceProfile` and `ConstraintSet`:

* combines the create / edit inputs into a single input
* removes the `edit` and `create` "command" inputs in favor of checking the current state of the science configuration and applying the updates accordingly


swalker2m:
Superseded by #513

PR #513, opened 2022-01-31
Science Configuration and Target Environment

swalker2m:
Makes the science configuration and target environment consistent with source profile and constraint set editing:

* One input for creation and editing with appropriate argument checking
* Removes explicit "edit" and "create" commands for science configuration
* Use "targetEnvironment" everywhere instead of a mix of "targetEnvironment" and "targets"


toddburnside (approved):
LGTM

PR #515, opened 2022-02-01
fix input names

swalker2m:
Fixes a few incorrectly named inputs.


modules/core/src/main/scala/lucuma/odb/api/schema/ScienceConfigurationMutation.scala
```diff
@@ -22,7 +22,7 @@ trait ScienceConfigurationMutation {
 
   implicit val InputObjectTypeGmosSouthLongSlit: InputObjectType[GmosSouthLongSlitInput] =
     InputObjectType[GmosSouthLongSlitInput](
-      "EditGmosSouthLongSlit",
+      "GmosSouthLongSlitInput",
```
cquiroz, line 25: Noted that one too :D

cquiroz (approved):
Thanks, LGTM

PR #519, opened 2022-02-07
Query ITC

swalker2m:
Adds an `itc` query to the observation which obtains ITC results for the observation when possible.  This is a work-in-progress and was needed for sequence step generation.  It will evolve as the ITC input schema evolves.

One change is to the Sangria resolver context, which was an `OdbRepo` and is now an `OdbCtx` containing a repo.  The `OdbCtx` has an `ItcClient` as well as the repo.  Along the way, some places in the schema implementation where a specific context was required have been changed so that `Any` context is acceptable, since it isn't used in those cases anyway.


modules/service/src/main/scala/lucuma/odb/api/service/Init.scala
```diff
@@ -45,7 +49,7 @@ object Init {
     "point": {
       "bandNormalized": {
         "sed": {
-          "galaxy": "SPIRAL"
+          "stellarLibrary": "O5_V"
```
swalker2m, line 52: This should really be a `"galaxy": "SPIRAL"` as it was before, but the ITC isn't currently accepting it and I wanted at least one observation to be defined sufficiently to generate ITC results.

modules/core/src/main/scala/lucuma/odb/api/schema/ConstraintSetSchema.scala
```diff
@@ -25,7 +24,7 @@ object ConstraintSetSchema {
   implicit val EnumTypeWaterVapor: EnumType[WaterVapor] =
     EnumType.fromEnumerated("WaterVapor", "Water vapor")
 
-  def AirMassRangeType[F[_]]: ObjectType[OdbRepo[F], AirmassRange] =
+  val AirMassRangeType: ObjectType[Any, AirmassRange] =
```
cquiroz, line 27: Why not just use `F[_]`?

modules/core/src/main/scala/lucuma/odb/api/schema/ObservationSchema.scala
```diff
@@ -144,18 +146,36 @@ object ObservationSchema {
...
           name        = "scienceConfiguration",
-          fieldType   = OptionType(ScienceConfigurationType[F]),
+          fieldType   = OptionType(ScienceConfigurationType),
           description = Some("The science configuration"),
           arguments   = List(ArgumentIncludeDeleted),
           resolve     = c => c.value.scienceConfiguration
         ),
 
+        Field(
+          name        = "itc",
```
cquiroz, line 156: Are you storing the itc results as part of the observation, Could this be provided by explore?

cquiroz:
Explore does get the itc results from ITC and could pass them as part of the selected mode. It seems instead odb is directly querying itc. Or is this for cli users?

modules/core/src/main/scala/lucuma/odb/api/schema/ConstraintSetSchema.scala
swalker2m: The context object is not used so it seemed better to allow any context than to require an `OdbCtx[F]` or `OdbRepo[F]`.

modules/core/src/main/scala/lucuma/odb/api/schema/ObservationSchema.scala
swalker2m: I thought it should be available as part of the API and at any rate, the output is needed in order to generate the sequence so I thought we might as well expose it as an intermediate step.  It seems like it is useful for API users.

I'm not planning on storing it per se, but I think I will stick it in an LRU cache (perhaps expiring after a short while?) or something since it is kind of an expensive thing to compute.  The `ItcSpectroscopyInput` seems like a perfect hash key.

cquiroz:
I think we may want to eventually move the client to the itc project but there is no hurry

cquiroz:
Unfortunately this is a moving target, we'll need to update this soon

cquiroz (approved):
LGTM

swalker2m:
> I think we may want to eventually move the client to the itc project but there is no hurry

That makes sense to me.  It does seem like every service should provide a corresponding client.  Perhaps even the ODB client used by explore?

swalker2m:
> Unfortunately this is a moving target, we'll need to update this soon

Yeah, I'm expecting that but wanted to have a start to build on now.

PR #521, opened 2022-02-08
Respect includeDeleted flag in target grouping queries

swalker2m:
@toddburnside discovered that the `includeDeleted` flag was being ignored in the target grouping queries. This PR will now treat deleted targets as nonexistent when `includeDeleted` is `false`.


toddburnside (approved):
LGTM. Thanks for fixing this.

PR #522, opened 2022-02-09
ITC result cache

swalker2m:
Updates the ITC client to use a simple forever `Map` cache of results.  I think this is good enough for the throwaway API service, especially since it is restarted daily.  There is a `useCache` argument (`true` by default) that can be used to skip the cache lookup.

Switches to the production ITC because I need to update the client for the updated ITC API.


cquiroz (approved):
Looks good
I'm hopping to make a cache on the itc side in the near future but make it persistent via e.g. redis or memcachd

PR #526, opened 2022-02-09
Update ItcSpectroscopyInput

swalker2m:
* Updates `ItcSpectroscopyInput` to match the current API.  Defines a bunch of encoders that might be better placed in `core` I don't know.
* Switches back to itc-staging by default
* Switches init back to spiral galaxy SED since the server should be able to swallow it now


cquiroz (approved):
LGTM

PR #534, opened 2022-02-18
Simplify state

swalker2m:
Removes the dependency on `cats-mtl` that I introduced to solve a problem that is more easily solved with a straightforward application of `StateT`.  Also removes unused sharing code.

This is a largish refactoring but **it doesn't change the schema / API** per se.  In some cases where multiple errors were present in a request, instead of reporting them all only the first may be reported.  In practice I don't think it's a big issue though.


toddburnside (approved):
I didn't read thoroughly, but it LGTM.

PR #548, opened 2022-03-15
Switch to UUID for Atom/Step, update Observe <-> ODB API.

swalker2m:
## Observe / ODB API Updates

This is a refactoring of the Observe / ODB interaction.  The main change is that Observe is now responsible for sharing with the ODB the actual static and dynamic step configurations that it is executing.

Apologies for an un-reviewable PR.  I've tried to make the description (below) reasonably complete though.

### Major Updates

* Switching Atom and Step IDs to typed-UUIDs (i.e., with a type letter prefix à la `Gid`).  I've called this grand innovation a `Uid`. `a-f1006c8d-659d-44a4-a557-79879c3e8f95` for example signifies an atom.  As part of this, in the manual sequence atoms and steps are stored directly with the observation instead of being referenced but housed in atom and step "tables".  The "dereferencing" code is accordingly deleted.

* Visits.  This is a concept carried over from the `ocs` codebase.  When all or part of an observation is executed by Observe there is a corresponding "visit".  There can be multiple visits, even on separate nights, before an observation is complete.  Visits have `Uid`s prefixed with `v`.  From the API perspective, they contain all the events, steps, and datasets that were produced during that visit.  Note, atoms disappear after execution in that they aren't included in the list of recorded steps (i.e., they are a scheduling concept only).  Problematic?

* `VisitRecord` and `StepRecord`.  Before Observe starts a visit, it uses the API to record the static configuration, adding a new `VisitRecord`.  The return value includes a visit ID that must be specified for subsequent events.  The ODB will check that the ID refers to a valid visit and reject the event otherwise.

```
# Record a new visit
recordGmosNorthVisit(
  # VisitRecord creation parameters
  input: GmosNorthVisitRecordInput!
): GmosNorthVisitRecord!

# Input parameters for creating a new GmosNorth VisitRecord
input GmosNorthVisitRecordInput {
  observationId: ObservationId!
  static:        GmosNorthStaticInput!
}

# A GmosNorth visit as recorded by Observe
type GmosNorthVisitRecord {
  # Visit id
  id: VisitId!

  # Created by Observe at time
  created: Instant!

  # Started at time
  startTime: Instant

  # Ended at time
  endTime: Instant

  # Step duration
  duration: Duration!

  # GmosNorth static instrument configuration
  staticConfig: GmosNorthStatic!

  # GmosNorth recorded steps
  steps(
    # Retrieve `first` values after the given cursor
    first: Int

    # Retrieve values after the one associated with this cursor
    after: Cursor
  ): GmosNorthStepRecordConnection!

  # Sequence events associated with this visit
  sequenceEvents(
    # Retrieve `first` values after the given cursor
    first: Int

    # Retrieve values after the one associated with this cursor
    after: Cursor
  ): SequenceEventConnection!
}


addSequenceEvent(
  # Sequence event description
  input: AddSequenceEventInput!
): SequenceEvent!

# SequenceEvent creation parameters
input AddSequenceEventInput {
  observationId: ObservationId!
  visitId:       VisitId!
  command:       SequenceCommand!
}
```

Similarly before each step is executed Observe must record the dynamic step configuration, creating a new `StepRecord`.  Steps also have IDs and these are required for subsequent step and dataset events.

```
# Record a new step
recordGmosNorthStep(
  # GmosNorth step configuration parameters
  input: GmosNorthStepRecordInput!
): GmosNorthStepRecord!

# Input parameters for creating a new GmosNorth StepRecord
input GmosNorthStepRecordInput {
  observationId: ObservationId!
  visitId:       VisitId!
  stepConfig:    GmosNorthDynamicInput!
}

# A GmosNorth step configuration as recorded by Observe
type GmosNorthStepRecord {
  # Step id
  id: StepId!

  # Visit id
  visitId: VisitId!

  # Created by Observe at time
  created: Instant!

  # Started at time
  startTime: Instant

  # Ended at time
  endTime: Instant

  # Step duration
  duration: Duration!

  # GmosNorth configuration for this step
  instrumentConfig: GmosNorthDynamic!

  # The executed step itself
  stepConfig: StepConfig!

  # Step events associated with this step
  stepEvents(
    # Retrieve `first` values after the given cursor
    first: Int

    # Retrieve values after the one associated with this cursor
    after: Cursor
  ): StepEventConnection!

  # Dataset events associated with this step
  datasetEvents(
    # Retrieve `first` values after the given cursor
    first: Int

    # Retrieve values after the one associated with this cursor
    after: Cursor
  ): DatasetEventConnection!

  # Datasets associated with this step
  datasets(
    # Retrieve `first` values after the given cursor
    first: Int

    # Retrieve values after the one associated with this cursor
    after: Cursor
  ): DatasetConnection!
}

addStepEvent(
  # Step event description
  input: AddStepEventInput!
): StepEvent!

# StepEvent creation parameters
input AddStepEventInput {
  observationId: ObservationId!
  visitId:       VisitId!
  stepId:        StepId!
  sequenceType:  SequenceType!
  stage:         StepStage!
}
```

* Adds `Input` types for GMOS static and dynamic step config that could, with just a bit more schema effort, be used to create a manual sequence via the API.  A bit more effort on top of that will make them editable.

### Minor Updates

* Removes the `generated` timestamp which recorded the client (Observe) creation time of events.

* Event IDs continue to be `Gid`s but it is no longer possible for the client to supply them.  Client-supplied ids are useful for UIs but it's unclear why Observe would supply them for events.

* Removes `Nanoseconds` units from the duration options.

* Introduces an `fpu` level in the GMOS sequence which contains either `customMask` or `builtin`.  Technically it contains both but only one will be set.


cquiroz:
Is it so that steps are not recorded anywhere until they are "visited"? Would observe call the api to get a snapshot of the steps?

Maybe we could document the interactions with observe?

swalker2m:
> Is it so that steps are not recorded anywhere until they are "visited"? 

That was my hope, yes. It may be that we need to cache them for the scheduler.

> Would observe call the api to get a snapshot of the steps?

Yes, that part hasn't changed from before.

> Maybe we could document the interactions with observe?

That's a good idea.  I'll put together a `.md` doc.

swalker2m:
> Maybe we could document the interactions with observe?

Here is the general idea from the Observe perspective:  https://github.com/gemini-hlsw/lucuma-tmp-api/blob/12d2bf25c6922f5463cabadc7beb1481eb985f44/docs/ObserveOdbInteraction.md

@jluhrs and @cquiroz you're probably best situated to evaluate this idea.  Discussion from anybody is welcome though.

cquiroz:
Thanks for the documentation it makes it all clearer to me

docs/ObserveOdbInteraction.md
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+```
+
+Notice that the `static` configuration is separated from the steps themselves and that the `acquisition` and `science` steps are segregated.  Observe would start by executing the acquisition steps and once accomplished switch to the science steps.  Within a sequence, the `nextAtom` (if any) is the one that Observe should execute while the `possibleFuture` shows what we expect to happen for the remainder of the sequence.  Because the possible future is uncertain, after finishing the `nextAtom`, Observe would repeat the query to obtain the latest `nextAtom` and latest `possibleFuture`.  This would continue until no `nextAtom` is forthcoming or until the observer decides to stop the observation.
+
+## Starting a Visit
+
+Before executing an observation, Observe would record the start of a "visit".  This produces a `VisitRecord` in the database.  In theory the `static` configuration in the query above can be modified or ignored altogether so part of recording a visit is to pass the actual static configuration that is used.
+
```
jluhrs, line 72: Would a change in the observation configuration trigger a new "visit"? I think that would make sense, Observe can record a new "visit" to acknowledge the configuration change for the remaining steps.

swalker2m: Yes, if you change the static configuration a new visit is needed.

docs/ObserveOdbInteraction.md
```diff
@@ -0,0 +1,236 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Notice that the `static` configuration is separated from the steps themselves and that the `acquisition` and `science` steps are segregated.  Observe would start by executing the acquisition steps and once accomplished switch to the science steps.  Within a sequence, the `nextAtom` (if any) is the one that Observe should execute while the `possibleFuture` shows what we expect to happen for the remainder of the sequence.  Because the possible future is uncertain, after finishing the `nextAtom`, Observe would repeat the query to obtain the latest `nextAtom` and latest `possibleFuture`.  This would continue until no `nextAtom` is forthcoming or until the observer decides to stop the observation.
+
```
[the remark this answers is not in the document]
swalker2m: When the event signifying the end of the last step in an atom is sent and the mutation completes, it must be recoded in the database.  It's true that we cannot kick off an asynchronous process and return a status code or something to the caller.

In other words, if the caller gets a response to the event message and afterwords decides to ask for the `nextAtom` then the `nextAtom` calculation will take into account that event.

rpiaggio:
Forgive my intrusion into this part of the domain which I know very little about. Just raising the question: would it be convenient for tracing purposes to record the current `nextAtom` (and maybe even the `possibleFuture`) when a new step is created?

swalker2m:
> would it be convenient for tracing purposes to record the current nextAtom (and maybe even the possibleFuture) when a new step is created?

Could you elaborate a bit?  I'm not sure what you mean.

rpiaggio:
I was just elaborating on your comment of

> atoms disappear after execution in that they aren't included in the list of recorded steps (i.e., they are a scheduling concept only). Problematic?

I'm assuming things work the following way:
- The backend suggests a sequence (not considering manual sequences here).
- Observe uses the suggestion to perform the observation and records what it did by invoking the backend.

If there's ever an observation step that seems surprising, it might be desirable to have a way to see what the backend was suggesting at that time.

Also, I don't know if the observer can edit the suggested step before executing it (without switching to a manual sequence).

PR #552, opened 2022-03-23
Library Updates + Switch to lucuma-core ConstraintSet

swalker2m:
Removes the tmp-api `ConstraintSetModel` in favor of `lucuma-core` `ConstraintSet`. Switches the GraphQL fields from `airmassRange` to `airMass` and `hourAngleRange` to `hourAngle` to be consistent with `ConstraintSet`.


modules/core/src/main/scala/lucuma/itc/client/ItcSpectroscopyInput.scala
```diff
@@ -185,24 +185,26 @@ object ItcSpectroscopyInput {
...
       Json.obj(
         "imageQuality"    -> screaming(a.imageQuality),
         "cloudExtinction" -> screaming(a.cloudExtinction),
         "skyBackground"   -> screaming(a.skyBackground),
         "waterVapor"      -> screaming(a.waterVapor),
         "elevationRange"  -> (a.elevationRange match {
-          case AirmassRange(min, max)             =>
+          case ElevationRange.AirMass(min, max)             =>
             Json.obj(
+              // TODO: we should switch this to airMass to be consistent
```
swalker2m, line 198: @cquiroz just calling this to your attention.  I switched to `airMass` to be consistent with `lucuma-core` so we might want to switch here too?

PR #557, opened 2022-03-31
GMOS North Long Slit Sequence Generation (1)

swalker2m:
This is a work-in-progress that has gotten quite big enough for a PR.  At this point, a "pristine" sequence can be generated from inputs.  The remaining work is hooking it into the API itself, taking into account executed steps to compute the `nextAtom` and `possibleFuture`.

At this point, only GMOS North is being done.  I want to get a vertical slice through the problem before fully including GMOS South.


cquiroz (approved):
I'm not sure I can really review this but LGTM

PR #559, opened 2022-04-01
Observation subtitle (aka "detail")

swalker2m:
Following the discussion in [Shortcut](https://app.shortcut.com/lucuma/story/1209/add-user-supplied-observation-details#activity-1221), this PR renames the currently unused observation `name` field to `subtitle`.

It also adds a generated `title` field to the observation output that matches the value currently shown in Explore.  I don't expect Explore to actually use the `title` field but this makes it available for API users.


rpiaggio (approved):
LGTM

PR #565, opened 2022-04-05
Query ITC for all observation targets, take max exposure time

swalker2m:
Updates the ITC client to perform successive queries for all targets in an observation's asterism.


swalker2m:
Screen Shot 2022-04-05 at 14 39 11

I perform one query per target and then look for the maximum exposure time from the results.  The API supports sending multiple "modes" or instrument configurations and getting back a (corresponding?) array of results.  Would it make sense and would it save any time to instead send a _single_ instrument mode and multiple source profiles and radial velocities?

cquiroz:
When you say multtiple source profiles, do you mean multiple targets? For an asterism?

swalker2m:
> When you say multtiple source profiles, do you mean multiple targets? For an asterism?

Yes.  I'm not sure it would optimize anything but I thought I'd ask.  The idea would be to make a single call with a single `InstrumentMode` and an array of, essentially, `(SourceProfile, RadialVelocity)`.  We could return a single result with the max or an array of one result per target.

cquiroz:
The original mode is meant for the UI where you want to get results as quickly as possible without waiting too much. For a backend the priorities may be different. I think we can adapt to the needs of the client

cquiroz (approved):
LGTM

PR #568, opened 2022-04-12
GmosNorthLongSlit sequence generation

swalker2m:
This PR implements a first pass at sequence generation for GMOS North longslit.  It will prefer a manual sequence if defined and otherwise generate an automatic sequence provided all required inputs are present and that we get an exposure time and count answer from ITC.  It looks at execution events to determine completed steps and take them into account when suggesting the next atom and possible future.

Screen Shot 2022-04-12 at 14 49 02

Caveats:

* Only GMOS North at the moment
* No smart GCAL so there's a stand-in flat step
* I don't know how to determine the acquisition exposure time so it is fixed
* There's no way to mark a dataset "bad" at the moment and consequently bad steps are not regenerated
* Doesn't yet drop in reacquisition sequences after time passes


swalker2m:
I struggled mightily to make the code compact and understandable.  Unfortunately I feel that I wasn’t very successful in the end despite the effort.  I need to turn to OCS tasks for a while before coming back to add GMOS South.  Hopefully a bit of distance from the code will be beneficial.  If not, Rob will most certainly have better ideas when we implement this in the real ODB.

rpiaggio:
Awesome! I'll start adding this to Explore. Is there a preloaded observation with all required parameters?

swalker2m:
> Is there a preloaded observation with all required parameters?

Yes, `o-5` should work.

swalker2m:
Another caveat, the spatial and wavelength offset, along with any other eventually configurable "Advanced" properties are all fixed at the moment.  This should be quick to address though.

modules/core/src/main/scala/lucuma/gen/Generator.scala
```diff
@@ -3,50 +3,96 @@
...
-   * reached).
-   *
-   * @param observed a computation that decides when the observation is done
+   * Runs the Generator to make the execution config with its static config,
+   * acquisition sequence, and science sequence.
    */
-  def science[F[_]: Sync](
-    observed: F[Boolean]
-  ): Stream[F, AtomModel[StepModel[D]]]
+  def run[F[_]: Monad, S, D](
```
rpiaggio, line 55: Can this be a method in `Generator` ?

swalker2m: Yes, I wanted it to be there but was stymied.  How can I stipulate that `F[_]` have a `Monad` if I add move it to the instance?

rpiaggio (approved):
LGTM. Disclaimer: I didn't get into the deep GmosN generation logic.

modules/core/src/main/scala/lucuma/gen/Generator.scala
rpiaggio: We can have an `implicit F: Monad[F]` parameter in the method itself.

swalker2m: Gah, of course!  Thanks, done.

PR #569, opened 2022-04-20
Remove extra stepType field

swalker2m:
On request, this PR removes the `stepType` field from `GmosNorthStep` and `GmosSouthStep`.  It is still available in `stepConfig.stepType`.


rpiaggio (approved):
LGTM thanks

PR #573, opened 2022-04-21
Remove extra setupTime field

swalker2m:
As @rpiaggio pointed out, you can get the `setupTime` from the `plannedTime` so this is redundant at best.


rpiaggio (approved):
LGTM

PR #575, opened 2022-04-22
Rename GMOS grating

swalker2m:
Renames GMOS "disperser" to "grating".  Sorry for the hassle but it isn't just capriciousness but rather a suggestion/request from Andy.  There's a matching change to `core` but it can be folded in later.

### DIFF
--- a/modules/core/src/main/scala/lucuma/itc/client/ItcSpectroscopyInput.scala
+++ b/modules/core/src/main/scala/lucuma/itc/client/ItcSpectroscopyInput.scala
@@ -215,22 +215,22 @@ object ItcSpectroscopyInput {
       )
 
   implicit val EncoderScienceConfigurationModel: Encoder[ScienceConfigurationModel] = {
-    case ScienceConfigurationModel.Modes.GmosNorthLongSlit(filter, disperser, fpu, _) =>
+    case ScienceConfigurationModel.Modes.GmosNorthLongSlit(filter, grating, fpu, _) =>
       Json.obj(
         "gmosN" ->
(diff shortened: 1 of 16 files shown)


cquiroz:
Do we need to change ITC to match?

swalker2m:
> Do we need to change ITC to match?

Yeah we probably should.  And now that you mention it, I guess I should update the `ScienceConfigurationModel` as well.

modules/core/src/main/scala/lucuma/itc/client/ItcSpectroscopyInput.scala
```diff
@@ -215,22 +215,22 @@ object ItcSpectroscopyInput {
...
 
   implicit val EncoderScienceConfigurationModel: Encoder[ScienceConfigurationModel] = {
-    case ScienceConfigurationModel.Modes.GmosNorthLongSlit(filter, disperser, fpu, _) =>
+    case ScienceConfigurationModel.Modes.GmosNorthLongSlit(filter, grating, fpu, _) =>
       Json.obj(
         "gmosN" ->
           Json.fromFields(
             List(
-              "disperser"  -> screaming(disperser),
+              "disperser"  -> screaming(grating),
```
swalker2m, line 223: Leaving the input until ITC is updated.

rpiaggio (approved):
LGTM

cquiroz (approved):
LGTM, does it need core 0.32?

swalker2m:
> LGTM, does it need core 0.32?

No, should be independent.

PR #576, opened 2022-04-25
Remove slitWidth from longslit modes

swalker2m:
Removes the `slitWidth` field of the long slit modes.  The slit width is determined by the FPU. It isn't clear what we'd do with `slitWidth` and we shouldn't allow it to differ from FPU anyway.  Over in `ScienceRequirements` there is a `focalPlaneAngle` which I believe is used to hold the user's requirement for the slit width.


cquiroz (approved):
LGTM

PR #581, opened 2022-04-28
Science Mode

swalker2m:
Swaps the existing `ScienceConfigurationModel` for `ScienceMode`.  Adds basic and advanced configuration options for each mode (GMOS North and GMOS South Long Slit).  Not much is done with the advanced configuration at the moment that the sequence is generated.  That will be the subject of the next PR.


modules/core/src/main/scala/lucuma/gen/gmos/longslit/GmosNorthLongSlit.scala
```diff
@@ -138,35 +134,31 @@ object GmosNorthLongSlit {
...
-        new Wavelength(λ.toPicometers + Δ.to[PosInt, Picometer])
+    ): ScienceSteps[NorthDynamic] =
 
       eval {
         for {
           _  <- NorthDynamic.exposure      := exposureTime.value
-          _  <- NorthDynamic.xBin          := mode.fpu.xbin(sourceProfile, imageQuality, sampling)
-          _  <- NorthDynamic.yBin          := GmosYBinning.Two
+          _  <- NorthDynamic.xBin          := mode.xBin(sourceProfile, imageQuality, sampling)
+          _  <- NorthDynamic.yBin          := mode.yBin
```
swalker2m, line 149: Slight use of advanced configuration here.  If you provide an explicit X or Y binning value it will get used.  The next step will be to use all the available advanced configuration options.

cquiroz (approved):
LGTM

PR #585, opened 2022-05-02
Clone obs

swalker2m:
By request, adds a simple `cloneObservation` mutation.  The `existingObservationId` parameter is required and all others are optional.  The idea is that you get a clone (except for observation status which is reset to `NEW`) by default but can override any of the other values if desired.


rpiaggio (approved):
LGTM

PR #588, opened 2022-05-04
Datasets and Dataset QA

swalker2m:
Refactors a bit to explicitly store datasets instead of just deriving them from events.  Adds a top-level query for datasets and, more importantly, a mutation to set the QA state.  This will be used to influence the sequence generation.


cquiroz (approved):
LGTM

PR #592, opened 2022-05-06
Switch to java.time.Duration from FiniteDuration.

toddburnside:


modules/core/src/main/scala/lucuma/odb/api/model/time.scala
```diff
@@ -9,37 +9,36 @@ import eu.timepit.refined.api.Validate.Plain
...
     Validate.fromPredicate(
-      (d: FiniteDuration) => d.length >= 0L,
-      (d: FiniteDuration) => s"$d is not negative",
+      (d: Duration) => d.toNanos >= 0L,
+      (d: Duration) => s"$d is not negative",
       Not(Less(shapeless.nat._0))
     )
 
-  type NonNegativeFiniteDuration = FiniteDuration Refined NonNegative
+  type NonNegativeDuration = Duration Refined NonNegative
```
swalker2m, line 28: We discussed unifying the two instances that I mistakenly created.  I can do it in another PR.

swalker2m (approved):
Looks great, thanks for taking care of this. :+1:

rpiaggio (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/model/time.scala
toddburnside: Actually, this is the unified instance. 😄 What remains in `gen/package.scala` is syntax specifically for the `AcqExposureTime` and other 2 types in that file. But, now they make use of the refined type in this file instead of redefining it.

PR #595, opened 2022-05-10
Event API Standardization

swalker2m:
Refactors the event API, standardizing events for the sequence, step and dataset.  Each event is given a `location` identifying the associated element and a `payload` with actual event data.  This separates the two concerns and follows the recommended practice of nesting objects to facilitate API evolution.

Events are not added or directly used by Explore so hopefully this will not occasion any changes to the UI.  Instead, this is part of the the API intended for use by Observe as it executes an observation.


toddburnside (approved):
LGTM

PR #596, opened 2022-05-10
Bulk Edit Inputs

swalker2m:
Renamed the bulk edit mutations `bulkEdit*` for clarity and to match the corresponding `BulkEdit*Input`s.  Nested the selection into its own `BulkEditSelectInput` object.


toddburnside (approved):
LGTM

PR #599, opened 2022-05-12
Don't store planned time summary

swalker2m:
A small, nonbreaking, change to stop storing the planned time summary.  These are supposed to be computed and will change as you edit the observation.  For now we can just generate a stable, if bogus, value.


cquiroz (approved):
LGTM

PR #600, opened 2022-05-18
Edit, create and clone mutations (for observations and targets)

swalker2m:
Reworks observation and target edit, create and clone methods (program is next, pending Todd's upcoming proposal PR).

* Removes the client-suggested IDs
* All the edits are now bulk edits so the "bulk" prefix is dropped.
* `updateTarget` and `updateObservation` were renamed `editTarget` and `editObservation` to match the name of the input and be consistent with bulk editing asterisms.
* The old bulk edit constraint set, etc. is just subsumed into `editObservation`.  (The one exception is `editAsterism` which requires a special `patch` input.)

It may be easiest to skip the tedious details and just look at the test mutations at the bottom.  Nevertheless, here are a few pointers:

## Edit

Observation and target edit method have the expected argument.  For example:

```
editObservation(
  input: EditObservationInput!
): [Observation!]!
```

The input in both cases has a `select` and a `patch` field appropriate to the thing being edited.

```
input EditObservationInput {
  select: ObservationSelectInput!
  patch:  ObservationPropertiesInput!
}
```

Selection has fields to pick which observation(s) you want to apply the patch to.  

```
input ObservationSelectInput {
  programId:      ProgramId
  observationId:  ObservationId
  observationIds: [ObservationId!]
}
```

Here `programId` will pick all the program's observations, `observationId` will pick a single observation, and `observationIds` will pick the listed observations.  `observationId` is just a (minor) convenience if you want to select just one observation, which is expected to be the common case.  You could just as well have a one element list in `observationIds`.  Will this be confusing?

The patch field describes the edits.  Only the changes are specified, so all the fields are optional.  Some can be set to `null` to delete the corresponding value.  For example `subtitle = null` to remove the observation subtitle.

```
# Observation properties
input ObservationPropertiesInput {
  # Subtitle adds additional detail to the target-based observation title, and is both optional and nullable
  subtitle: NonEmptyString

  # The observation status will default to New if not specified when an observation is created and may be edited but not deleted
  status: ObsStatus

  # The observation active status will default to Active if not specified when an observation is created and may be edited but not deleted
  activeStatus: ObsActiveStatus

  # The targetEnvironment defaults to empty if not specified on creation, and may be edited but not deleted
  targetEnvironment: TargetEnvironmentInput

  # The constraintSet defaults to standard values if not specified on creation, and may be edited but not deleted
  constraintSet: ConstraintSetInput

  # The scienceRequirements defaults to spectroscopy if not specified on creation, and may be edited but not deleted
  scienceRequirements: ScienceRequirementsInput

  # The scienceMode describes the chosen observing mode and instrument, is optional and may be deleted
  scienceMode: ScienceModeInput

  # Whether the observation is considered deleted (defaults to PRESENT) but may be edited
  existence: Existence
}
```

## Create

Creation reuses the patch input and interprets it differently to check that all fields needed for creation are defined.

```
createObservation(
  # Observation description
  input: CreateObservationInput!
): Observation!

# Observation creation parameters
input CreateObservationInput {
  programId:  ProgramId!
  properties: ObservationPropertiesInput
}
```

## Clone

Clone is similar but you clone one observation/target at a time.  There is an optional `patch` input to perform edits on the new cloned item.

```
cloneObservation(
  # Parameters for cloning an existing observation
  input: CloneObservationInput!
): Observation!

input CloneObservationInput {
  observationId: ObservationId!
  patch:         ObservationPropertiesInput
}
```

### DIFF
--- a/modules/service/src/test/scala/test/targets/TargetCloneSuite.scala
+++ b/modules/service/src/test/scala/test/targets/TargetCloneSuite.scala
@@ -4,6 +4,7 @@
 package test
 package targets
 
+import cats.syntax.option._
 import io.circe.literal._
 
 class TargetCloneSuite extends OdbSuite {
@@ -11,9 +12,8 @@ class TargetCloneSuite extends OdbSuite {
   // Clone an existing deleted, target.
   queryTest(
     query ="""
-      mutation CloneTarget {
-        cloneTarget(existingTargetId: "t-4", suggestedCloneId: "t-abc") {
-          id
+      mutation CloneTarget($cloneInput: CloneTargetInput!) {
+        cloneTarget(input: $cloneInput) {
           properties {
             name
           }
@@ -24,24 +24,34 @@ class TargetCloneSuite extends OdbSuite {
     expected =json"""
       {
         "cloneTarget": {
-          "id": "t-abc",
           "properties": {
-            "name": "NGC 3312"
+            "name": "Biff"
           },
           "existence": "PRESENT"
         }
       }
     """,
-    None,
+    variables = json"""
+      {
+        "cloneInput": {
+          "targetId": "t-4",
+          "patch": {
+            "properties": {
+              "name": "Biff"
+            }
+          }
+        }
+      }
+    """.some,
     clients = List(ClientOption.Http)
   )
 
+  //clexistingTargetId: "t-4", suggestedCloneId: "t-a", observationIds: [ "o-3", "o-4" ]) {
   // Clone an existing target and replace it in observations 3 and 4
   queryTest(
     query ="""
-      mutation CloneAndReplaceTarget {
-        cloneTarget(existingTargetId: "t-4", suggestedCloneId: "t-a", observationIds: [ "o-3", "o-4" ]) {
-          id
+      mutation CloneAndReplaceTarget($cloneInput: CloneTargetInput!) {
+        cloneTarget(input: $cloneInput) {
           properties {
             name
           }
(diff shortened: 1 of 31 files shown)


modules/service/src/test/scala/test/targets/TargetCloneSuite.scala
```diff
@@ -22,35 +22,52 @@ class TargetCloneSuite extends OdbSuite {
...
+          "patch": {
+            "name": "Biff"
+          }
+        }
+      }
+    """.some,
     clients = List(ClientOption.Http)
   )
 
+  //clexistingTargetId: "t-4", suggestedCloneId: "t-a", observationIds: [ "o-3", "o-4" ]) {
```
toddburnside, line 43: remove commented code?

toddburnside (approved):
LGTM

modules/service/src/test/scala/test/targets/TargetCloneSuite.scala
swalker2m: Thanks. :+1:

cquiroz:
Maybe having both `observationId` and `observationIds` could be confusing? I wonder if API users would try to do stuff like putting the obs id twice

cquiroz:
Reusing the patch seems good, it should reduce the size of the schema

cquiroz (approved):
I'll assume the code is correct! LGTM

rpiaggio:
> Maybe having both `observationId` and `observationIds` could be confusing? I wonder if API users would try to do stuff like putting the obs id twice

I agree with this.

rpiaggio (approved):
LGTM. The API feels more compact and harmonious.

swalker2m:
> Maybe having both observationId and observationIds could be confusing? I wonder if API users would try to do stuff like putting the obs id twice

That wouldn't cause any error, since it just takes the union of all the selection options, but I was worried it might be confusing too.  Removing `observationId` singular will mean having to specify a single observation with list syntax, e.g., `[ "o-2" ]` but I'll do it.  If Andy feels like it would be useful to have later it's easy to add.

PR #601, opened 2022-05-18
somewhat better ITC query failure messages

swalker2m:
Instead of an "internal server error" for ITC issues, we can display a slightly more informative message.

```
{
  "data": {
    "observation": {
      "execution": {
        "executionConfig": null
      }
    }
  },
  "errors": [
    {
      "message": "Could not extract required ITC query input from o-2",
      "path": [
        "observation",
        "execution",
        "executionConfig"
      ],
      "locations": [
        {
          "line": 10,
          "column": 7
        }
      ]
    }
  ]
}
```

I'm not sure whether this is the right thing to do -- it isn't like the user made some mistake that they can fix -- but it seems better than 500 internal server error?   The execution config result is optional, should we instead just silently return none?


swalker2m:
For context, this was an observation with a "gaussian" source profile.

cquiroz (approved):
LGTM

PR #605, opened 2022-05-23
More refined, less fromLong

swalker2m:
A couple of hopefully minor API changes, despite all the LOCs deleted to achieve them:

1. Defines and uses a couple of refined scalars to better document the API.  For example, duration values like exposure time cannot be negative now.
2. Removes the `fromLong` and `fromDecimal` options when querying for outputs and defining numeric value inputs.  For example, you can specify an angle as microarcseconds, etc. as before

```
input AngleInput {
  microarcseconds: Long
  microseconds: BigDecimal
  milliarcseconds: BigDecimal
  milliseconds: BigDecimal
  arcseconds: BigDecimal
  seconds: BigDecimal
  arcminutes: BigDecimal
  minutes: BigDecimal
  degrees: BigDecimal
  hours: BigDecimal
  dms: String
  hms: String
}

type Angle {
  # Angle in µas
  microarcseconds: Long!

  # Angle in µs
  microseconds: BigDecimal!

  # Angle in mas
  milliarcseconds: BigDecimal!

  # Angle in ms
  milliseconds: BigDecimal!

  # Angle in asec
  arcseconds: BigDecimal!

  # Angle in sec
  seconds: BigDecimal!

  # Angle in amin
  arcminutes: BigDecimal!

  # Angle in min
  minutes: BigDecimal!

  # Angle in deg
  degrees: BigDecimal!

  # Angle in hrs
  hours: BigDecimal!
}
```

 but the `"fromLong": { "units": "MICROARCSECONDS", "value": 12345 }` option would be removed.  I think these were not and would never be used and only added weight and complexity.  Correct me if I'm wrong!


toddburnside:
> but the `"fromLong": { "units": "MICROARCSECONDS", "value": 12345 }` option would be removed. I think these were not and would never be used and only added weight and complexity. Correct me if I'm wrong!

Excellent. I don't think these were actually in the graphql api and I was confused as to why they were there. 😄

cquiroz:
I agree `fromLong` is likely not needed. Perhaps it was used by cli users? I doubt it though

cquiroz (approved):
LGTM

swalker2m:
> Perhaps it was used by cli users? I doubt it though

It was early on and I think I was just exploring how to do values with units.  Instead of deciding it had both approaches but I think in the end it's better to decide on one way.

swalker2m:
I'm going to merge this because I want it for the advanced config exposure mode.

PR #609, opened 2022-05-25
More advanced gmos long slit configuration

swalker2m:
Adds `overrideWavelength` and `overrideExposureTimeMode` to the science mode advanced configuration input and output.  They can be set to ignore the science requirements and use a different wavelength and/or exposure time mode.  If set to explicit exposure count and time then the call to ITC is bypassed (for now anyway -- eventually we need to make a call to get the acquisition time).

Screen Shot 2022-05-25 at 16 29 29

Screen Shot 2022-05-25 at 16 28 11


toddburnside (approved):
LGTM

PR #610, opened 2022-05-25
Add Proposals

toddburnside:
I still need to add some tests for the `updateProgram` and `createProgram` mutations to test the graphQL stuff. But, I'd like to take care of this part first to avoid more merge conflicts.


modules/core/src/main/scala/lucuma/odb/api/model/gmos/longslit/GmosLongslitMath.scala
```diff
@@ -98,7 +98,7 @@ object GmosNorthLongslitMath extends GmosSiteSpecificLongslitMath {
   override def site: Site = Site.GN
 
   def effectiveSlitWidth(fpu: GmosNorthFpu, p: SourceProfile, iq: ImageQuality): Angle =
-    fpu.slitWidth.getOrElse(IfuSlitWidth) min effectiveSize(p, iq)
```
toddburnside, line 101: I'm not sure if the changes in this file are correct. There were changes in lucuma-core that broke this code, and these changes compiled...

swalker2m: Yes thanks, that looks right to me.

modules/service/src/test/scala/QuerySuite.scala
```diff
@@ -14,6 +14,32 @@ class QuerySuite extends OdbSuite {
...
+              title
+              proposalClass {
+                __typename
+                minPercentTime
+                ... on LargeProgram {
+                  minPercentTotalTime
+                  totalTime {
+                    seconds
+                  }
+                }
```
swalker2m, line 27: In other parts of the API where there are multiple options (e.g., source profile point, uniform and gaussian and target sidereal and nonsidereal tracking), the pattern is to put all the fields as optional.  There are benefits and drawbacks to that approach but I've tried to make it consistent.  This would be a departure, which may be fine.  Maybe the sheer number of options makes it more desirable to do it this way?

swalker2m (approved):
This looks great to me.  Thanks for taking care of the `NonNegDuration` thing and updating the longslit math.  :+1:

modules/service/src/test/scala/QuerySuite.scala
toddburnside: I struggled a bit with this. In the temporary implementation in Explore, the proposal model was flatter, with an enumeration for the Proposal Class and the `minPercentTime`, `minPercentTotalTime` and `totalTime` fields directly on the proposal. But, with the latter 2 fields only being applicable to a small subset of the proposal classes, it seemed it would be confusing as to when they were required and when they were forbidden. From the UI perspective that isn't too bad (and is probably easier), but it seemed less than ideal for API users. 
If we want to change this, I'm not irrevocably tied to this approach.

modules/core/src/main/scala/lucuma/odb/api/schema/ProgramMutation.scala
```diff
@@ -18,14 +20,150 @@ trait ProgramMutation {
...
+      ReplaceInputField("minPercentTotalTime", IntPercentType.createRequiredEditOptional("minPercentTotalTime", "intensive")),
+      ReplaceInputField("totalTime",           InputObjectTypeNonNegDuration.createRequiredEditOptional("totalTime", "intensive"))
+    )
+
+  implicit val InputObjectTypeProposalClassInput: InputObjectType[ProposalClassInput] =
+    InputObjectType[ProposalClassInput](
+      "ProposalClassInput",
+      "Proposal class. Choose exactly one class type",
+      List(
+        InputField("classical",          OptionInputType(InputObjectTypeClassicalInput),          "classical"),
```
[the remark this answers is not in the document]
toddburnside: Nice.

modules/service/src/test/scala/QuerySuite.scala
toddburnside: Hmmm. Yeah. I see. OK. I'll merge and we'll ponder it as we go forward. The ability to cut and paste from output to input is kind of nice.

swalker2m:
>  I'll merge and we'll ponder it as we go forward.

:+1:

PR #612, opened 2022-05-26
normalize (un)delete

swalker2m:
Updates `deleteObservation`, `deleteTarget`, `undeleteObservation` and `undeleteTarget` to follow the convention of taking a single input object parameter named after the mutation itself (e.g., `DeleteObservationInput`).  All the inputs reuse the same selection input used by the corresponding `edit` method.  For example,

```
{
  "deleteObservationInput": {
    "select": {
       "observationIds": [ "o-2" ]
    }
  }
}
```

For better or worse, this makes it very easy to delete many observations at once.  For example by selecting the program you could delete _all_ of its observations in a single mutation.  Fortunately this is easy to undo using the same selection input with `undeleteObservation`.

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -140,17 +140,7 @@ object ObservationModel extends ObservationOptics {
   object PropertiesInput {
 
     val Empty: PropertiesInput =
-      PropertiesInput(
-        subtitle            = Input.ignore,
-        status              = Input.ignore,
-        activeStatus        = Input.ignore,
-        targetEnvironment   = Input.ignore,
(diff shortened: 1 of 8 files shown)


swalker2m:
By the way, there's nothing in `deleteObservation` etc. that can't be done with a normal `editObservation` etc.  In fact it's implemented using a normal edit.  Maybe we don't need specialized mutations for them at all?  It's not likely a frequent operation, I guess.

toddburnside:
> By the way, there's nothing in `deleteObservation` etc. that can't be done with a normal `editObservation` etc. In fact it's implemented using a normal edit. Maybe we don't need specialized mutations for them at all? It's not likely a frequent operation, I guess.

I guess it makes the intent clearer, even if not strictly necessary. API uses may not think much about the existence flag.

toddburnside (approved):
LGTM

PR #614, opened 2022-05-31
New Observation Properties

swalker2m:
Adds two new observation properties and renames another.

* Add `visualizationTime`, which is used to compute the average parallactic position angle, proper-motion-adjusted coordinates and eventually non-sidereal target coordinates
* Add `posAngleConstraint` which is used to limit which values may be selected for the position angle
* Rename `config` to `manualConfig` to contrast it with the automatic sequence generation

The `PosAngleConstraint` differs from the core version because I found it difficult to map to the API.  What I ended up with seems simple to me, but I'm not sure and am open to suggestions.  To get a sense for it, skip down to the bottom where `MutationSuite` defines a few example mutations and results.

`PosAngleConstraint` is optional and can be deleted by setting it to `null`.  If not present, the the position angle is not constrained (i.e., `Unconstrained` in the core model).  Setting it to `AVERAGE_PARALLACTIC` and yet assigning a position angle results in a `ParallacticOverride` in the core model.  I think we can leave the core model alone and just write a codec to  do the translation.


cquiroz:
Where would the code live? in `lucuma-schema`?

swalker2m:
> Where would the code live? in `lucuma-schema`?

For the codec?  Good question.  If that makes sense I can write it and add it there.

modules/service/src/test/scala/MutationSuite.scala
```diff
@@ -261,5 +261,170 @@ class MutationSuite extends OdbSuite {
...
+          }
+        }
+      }
+    """,
+    expected = json"""
+      {
+        "editObservation" : [
+          {
+            "id": "o-3",
+            "posAngleConstraint": null
```
cquiroz, line 329: This would be `Unconstrained` I guess

swalker2m: Exactly.

cquiroz (approved):
LGTM

PR #618, opened 2022-06-06
Payload

swalker2m:
Adds a `Result` type wrapper around all mutation results.  All mutations now follow the pattern:

1. Verb name `doFoo`
2. Accepting a `DoFooInput!`
3. Returning a `DoFooResult`

At this point for the most part the result just has a field that is the same as the previous return value.  Clone mutations now include the original and the new (possibly edited) clone.

Since several of the mutations work on multiple objects at once, they were pluralized. For example, instead of `editObservation` it is `editObservations` with an `EditObservationsInput!` and an `EditObservationsResult`.


toddburnside (approved):
Looks good. But smells like work for Explore. 😉

PR #621, opened 2022-06-10
Nix Paging WIP

swalker2m:
This is a work-in-progress on transforming to a paging-free API.  It's easy to remove paging and instead return a list of results, but I wanted to offer at least a hint at ways that we might filter those results.  Looking at other APIs there are a couple of approaches.  One is to write a DSL query string, [like Shopify does](https://shopify.dev/api/usage/search-syntax).  For example

```graphql
{
  products(first: 5, query: "(title:Caramel Apple) OR (inventory_total:>500 inventory_total:<=1000)" ) {
    ...
  }
}
```

A drawback would be that there's nothing to guide a user in writing these search strings, though perhaps they are easy enough to figure out?  I'm skeptical of introducing a DSL, but would listen to anybody who thinks this is a good approach.

Another approach is [adding custom directives](https://dgraph.io/docs/graphql/queries/search-filtering/), but that seems complicated and I didn't seriously consider it.

[Hasura](https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/query-filters/), generates an API that closely follows SQL.  I'm not sure how intuitive this is for non-programmers but it seemed straightforward enough to me so I adopted a poor-man's query filtering based on it.

This draft PR partially transforms a single query, `programs`.  If we were to adopt this then I think all the mutation `select` objects should probably accept the same `WHERE` filters.

Screen Shot 2022-06-10 at 15 55 27

`WHERE` and `LIMIT` are capitalized, along with everything else SQL-ish to make it stand apart from the fields in the actual objects like `title` or TAC `category`.  There's a query results wrapper, `ProgramSelectResult`, that holds the actual matching programs, a count of all the matches even beyond those in the capped `matches` list and a `hasMore` boolean that lets you know whether there are additional items you're not getting.

Screen Shot 2022-06-10 at 13 08 40

`WhereProgram` can have nested `AND`, `OR` or `NOT` filters and the nesting is unlimited.  I think we can cap the depth of queries though and we should probably do that to protect ourselves.   By default if you specify multiple items then they must all match.  For example,

```json
{
  "whereProgram": {
    "name": {
      "LIKE": "%dark%friends%"
    },
    "id": {
      "IN": [ "p-2", "p-3", "p-4" ]
    }
  }
}
```

requires the name to match the pattern _and_ for the ID to be one of those provided.

To do this in a type-safe way requires creating a lot of type-specific input objects.  For example, the TAC category matcher shouldn't match on just any string but on real TAC categories.

Screen Shot 2022-06-10 at 13 17 33

Here is a silly `WHERE` that puts some of these things together

```json
{
  "whereProgram": {
    "name": { "LIKE": "%dark%friends%" },
    "proposal": { 
      "OR": [
        { "IS_NULL": true },
        {
          "category": {
            "EQ": "SMALL_BODIES"
          }
        }
      ]
    }
  }
```

This will match programs whose name is defined and matching the given `LIKE` pattern while also either having no proposal or a proposal with TAC category `SMALL_BODIES`.

Does anybody have better ideas?  Is this too dumb?  The query fields and definitions are wide open in Sangria at least. For example, I could add a `text` field that doesn't exist anywhere in the model and instead matches on the program name, proposal title and abstract. Would this be hard to implement with Grackle?

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ProposalInput.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ProposalInput.scala
@@ -20,23 +20,27 @@ import lucuma.odb.api.model.EitherInput
 import lucuma.odb.api.model.syntax.input._
 import lucuma.odb.api.model.syntax.lens._
 import lucuma.odb.api.model.syntax.validatedinput._
-import monocle.Focus
+import monocle.{Focus, Lens}
 
 import scala.collection.immutable.SortedMap
-
 import ProposalInput._
 
 final case class PartnerSplit(partner: Partner, percent: IntPercent)
 
 object PartnerSplit {
-  val partner = Focus[PartnerSplit](_.partner)
-  val percent = Focus[PartnerSplit](_.percent)
+
+  val partner: Lens[PartnerSplit, Partner] =
+    Focus[PartnerSplit](_.partner)
+
+  val percent: Lens[PartnerSplit, IntPercent] =
+    Focus[PartnerSplit](_.percent)
 
   implicit val decoderPartnerSplit: Decoder[PartnerSplit] = deriveDecoder[PartnerSplit]
 
   implicit val eqPartnerSplit: Eq[PartnerSplit] = Eq.instance {
     case (PartnerSplit(a1, b1), PartnerSplit(a2, b2)) => a1 === a2 && b1 === b2
   }
+
 }
 
 final case class ProposalInput(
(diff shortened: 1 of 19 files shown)


rpiaggio:
Looks good and gives all the flexibility everyone will ever need.

@tpolecat can all of this be generated automatically by grackle, like basura does?

BTW in the schema above, is `totalCount` meant to be an `Int!`?

tpolecat:
Hi, this looks pretty good I think, and Grackle can already do most of this.

There is a general filtering structure in the internal Query AST that includes all the predicates you mention here, I think.. The `Predicate` structure includes the following at the moment:

- `TRUE`, `FALSE`
- `And` , `Or`, `Not`
- `Eql`, `NEql`
- `Contains`
- `Lt`, `LtEql`, `Gt`, `GtEql`
- `IsNull`
- `In`
- `Matches` (takes a regex)
- `StartsWith`

There are also facilities for doing `Order`, `Offset`, and `Limit`, as well as `Count` (for your `totalCount`). There is no current facility for `hasMore` but we could hack this up by always selecting one more than the specified `Limit`. 

So :+1: I think your proposal is good and I see a clear way to implement it.

By the way if we're allowing `LIMIT` it might make sense to also support `OFFSET`. This would let API users perform paging if they really really want to fetch a huge data set ... they just have to spread it over multiple requests.

tpolecat:
> Yes, thanks. Fixed. I wonder if we would want to keep that anyway since it will require continuing to select rows even after the `LIMIT` cap is reached.

The way to do this is select `LIMIT + 1` internally and throw out the last row, but note the fact that it exists so you know `hasMore` is true. So it ends up not really costing anything. (Not implemented yet in Grackle.)

swalker2m:
Thanks @tpolecat and @rpiaggio for all the feedback!  I'm going to close this and keep working on it.

PR #622, opened 2022-06-10
Wavelength dithers decimal nm

swalker2m:
Switches to decimal wavelength dithers in advanced config, adds `Nm` to the name to clarify the units, and fixes a bug in creating the advanced config input type.


rpiaggio (approved):
LGTM

PR #634, opened 2022-06-28
Remove Paging

swalker2m:
Removes paging throughout the API, replacing it with `WHERE` clause filtering, `OFFSET` indexing, and `LIMIT` along the lines presented in #621.  For example,

```
query Programs {
  programs(WHERE: { name: { LIKE: "%Jupiter%"  }}, OFFSET: "p-2", LIMIT: 100 ) {
    matches {
      id
      name
    }
    hasMore
  }
}
```

The return type includes the matching items in `matches` and whether there were additional results that were not returned (`hasMore`).

At the risk of mutiny, I also updated the edit methods.  For example, the old pattern was along the lines of:

```
mutation EditObservations($editObservationsInput: EditObservationsInput!) {
  editObservations(input: $editObservationsInput) {
    observations {
      id
    }
  }
}

{
  "editObservations": {
    "select": {
      "observationIds": [ "o-2, "o-3" ]
    },
    "patch": {
      "scienceMode": {
        "gmosSouthLongSlit": {
          "advanced": {
            "explicitWavelengthDithersNm": [ 5.1, 2.3, 3.1, 2 ]
          }
        }
      }
    }
  }
}
```

To match the SQL-ishness of the query API, these returned to `update*` (sorry).  The `select` field in the input was replaced with the more powerful `WHERE` clause from the queries and `patch` turned into `SET`.  For example,

```
mutation UpdateObservations($updateObservationsInput: UpdateObservationsInput!) {
  updateObservations(input: $updateObservationsInput) {
    observations {
      id
    }
  }
}

{
 "updateObservationsInput": {
    "SET": {
      "scienceMode": {
        "gmosSouthLongSlit": {
          "advanced": {
            "explicitWavelengthDithersNm": [ 5.1, 2.3, 3.1, 2 ]
          }
        }
      }
    },
    "WHERE": {
      "id": { "IN": [ "o-2", "o-3" ] }
    }
  }
}
```

Similarly, the creation `properties` and clone `patch` field turned into `SET` to match `update`.


cquiroz (approved):
LGTM. In god (aka Shane) we trust

toddburnside (approved):
It's hard to take it all in, but I looked through it and it seems OK to me.

PR #639, opened 2022-06-29
Remove source profile editing inconsistency

swalker2m:
The source profile input included a couple of options that made it more convenient for interactive API users to edit brightness values or emission lines.  Namely, instead of having to set the entire list of brightness values in one fell swoop you could `editBrightnesses` or `deleteBrightnesses` and leave unmentioned brightness values alone.  For example this would update (or add) a `B` brightness value and delete the `SLOAN_U` value leaving everything else the same:

```
{
  "updateTargetsInput": {
    "SET": {
      "sourceProfile": {
        "point": {
          "bandNormalized": {
            "editBrightnesses": [
              {
                "band": "B",
                "value": 13,
                "units": "VEGA_MAGNITUDE"
              }
            ],
            "deleteBrightnesses": [
              "SLOAN_U"
            ]
          }
        }
      }
    },
    "WHERE": {
      "id": { "EQ": "t-2" }
    }
  }
```

These options were not used by Explore and were intended to make human interactive use easier.  Unfortunately they are directives (like `SET`, `ADD`, `DELETE`, etc.) and work differently than any other normal editing fields in the API.  There is a similar capability for fine-grained editing asterisms, but that is done in a separate mutation, `updateAsterism`, which takes a list of `ADD` and `DELETE` commands.  If we want fine-grained brightness or emission lines editing I think we should similarly offer that in a dedicated mutation.

I'll skip it for now though and just remove `editBrightnesses` and `deleteBrightnesses` for consistency.  Does this sound reasonable?


swalker2m:
To be clear, now to simply edit a single brightness (or emissions line) value requires setting the entire list.  This is intuitive I suppose and perhaps not that much of a burden if you can just copy, paste and edit.

```
{
  "updateTargetsInput": {
    "SET": {
      "sourceProfile": {
        "point": {
          "bandNormalized": {
            "brightnesses": [
              {
                "band": "B",
                "value": 12.7,
                "units": "VEGA_MAGNITUDE"
              },
              {
                "band": "J",
                "value": 10.279,
                "units": "VEGA_MAGNITUDE",
                "error": 0.0009
              },
              {
                "band": "H",
                "value": 9.649,
                "units": "VEGA_MAGNITUDE",
                "error": 0.0120
              },
              {
                "band": "K",
                "value": 9.425,
                "units": "VEGA_MAGNITUDE",
                "error": 0.0170
              },
              {
                "band": "SLOAN_U",
                "value": 14.147,
                "units": "AB_MAGNITUDE",
                "error": 0.0050
              },
              {
                "band": "SLOAN_G",
                "value": 12.924,
                "units": "AB_MAGNITUDE",
                "error": 0.0020
              },
              {
                "band": "SLOAN_R",
                "value": 12.252,
                "units": "AB_MAGNITUDE",
                "error": 0.0020
              },
              {
                "band": "SLOAN_I",
                "value": 11.888,
                "units": "AB_MAGNITUDE",
                "error": 0.0020
              },
              {
                "band": "SLOAN_Z",
                "value": 11.636,
                "units": "AB_MAGNITUDE",
                "error": 0.0020
              }
            ]
          }
        }
      }
    },
    "WHERE": {
      "id": { "EQ": "t-2" }
    }
  }
}
```

toddburnside (approved):
LGTM

PR #640, opened 2022-06-29
simple transaction test

swalker2m:
A simple test that when editing multiple items a failure to any means nothing is changed, even if individually it would be successful for some items.


toddburnside (approved):
LGTM

PR #646, opened 2022-07-05
Respect advanced in itc call

swalker2m:
Updates the ITC call to respect the "override" exposure time mode and wavelength options in the science mode.


toddburnside (approved):
Excellent. Thanks for doing this.

PR #656, opened 2022-07-18
includeDeleted

swalker2m:
The PR adjusts the API to

* Move the `existence` property from the `Program`, `Observation`, and `Target` `WHERE` clauses to the update input itself, but expressed as a boolean `includeDeleted` flag.
* Remove the `includeDeleted` flag from simple one-value-by-id `program`, `observation` and `target` queries since presumably if you're searching for a particular id you want it whether or not it is deleted.
* Make `deleteObservation`, `undeleteObservation`, `deleteTarget` and `undeleteTarget` not offer an `includeDeleted` argument because it is now implied  by the nature of the operation itself.

Moving the `existence` flag was done to

* Provide just one way to indicate whether or not to include deleted values throughout the API.  Previously it was via the `includeDeleted` flag in some queries and via the `existence` property in `WHERE` clauses elsewhere.  Now it is consistently via the `includeDeleted` flag.
* Fix a bug that was introduced by trying to default the value of the `existence` property to remove deleted values.  This was problematic because of the possibility of nesting clauses with their own default `existence` properties.

## Example

Query all targets, including deleted values

```
      query AllTargets {
        targets(WHERE: { programId: { EQ: "p-2" } }, includeDeleted: true) {
          matches {
            id
            name
          }
        }
      }
```

Query a particular target, whether or not it is deleted.

```
      query OneTarget {
        target(targetId: "t-2") {
          id
          name
        }
      }
```

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/ObservationModel.scala
@@ -250,7 +250,8 @@ object ObservationModel extends ObservationOptics {
   final case class UpdateInput(
     SET:   PropertiesInput,
     WHERE: Option[WhereObservationInput],
-    LIMIT: Option[NonNegInt]
+    LIMIT: Option[NonNegInt],
+    includeDeleted: Boolean = false
   ) extends TopLevelUpdateInput[Observation.Id, ObservationModel] {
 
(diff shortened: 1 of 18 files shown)


toddburnside (approved):
LGTM. I'll try to get it incorporated into lucuma-schemas and explore soon.

PR #660, opened 2022-07-20
Give first obs in initial data a valid configuration

toddburnside:
NGC-5949 is a North only target, but it had a GMOSSouth configuration


swalker2m (approved):
:+1:

PR #661, opened 2022-07-21
Remove duplicate  enum

swalker2m:
The schema defined both `Partner` and `partner` enums and they were identical. This removes `partner` and moves some proposal stuff over from `ProgramSchema` over to where it belongs in `ProposalSchema`.

The only visible difference should be that `partner` goes away and `PartnerSplitsInput` `partner` field is now type `Partner` instead of `partner`.

Screen Shot 2022-07-21 at 15 03 12
Screen Shot 2022-07-21 at 15 01 28


toddburnside (approved):
LGTM

PR #703, opened 2022-12-11
Set a default SED

cquiroz:
I did a naive fix to support empty SED but it was too naive
Turns out we can't update `lucuma-core` is it is scala-3 only now. Given this project is going to be short lived we may as well just assign a default SED until `lucuma-odb` proper can solve it
This is working with explore and the latest `lucuma-core`

### DIFF
--- a/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
+++ b/modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
@@ -303,10 +303,14 @@ object SourceProfileModel {
 
     import lucuma.odb.api.model.targetModel.SourceProfileModel.BandNormalizedInput._
 
-    override val create: ValidatedInput[BandNormalized[T]] =
-      (sed.notMissingAndThen("sed")(_.toUnnormalizedSed),
-       brightnesses.notMissingAndThen("brightnesses")(brightnessCreator[T])
-      ).mapN { (sed, bright) => BandNormalized(sed, bright) }
+    override val create: ValidatedInput[BandNormalized[T]] = {
+      // we can't update lucuma-core thus we need to default to somthing
+      val modelSed = sed.toOption.map(_.toUnnormalizedSed)
+        .getOrElse(UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.O5V).validNec)
+      (modelSed, brightnesses.notMissingAndThen("brightnesses")(brightnessCreator[T])).mapN { (sed, bright) =>
+        BandNormalized(sed, bright)
+      }
+    }
 
     override val edit: StateT[EitherInput, BandNormalized[T], Unit] =
       for {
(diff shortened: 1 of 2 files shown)


toddburnside (approved):
LGTM

modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
```diff
@@ -303,10 +303,14 @@ object SourceProfileModel {
...
-       brightnesses.notMissingAndThen("brightnesses")(brightnessCreator[T])
-      ).mapN { (sed, bright) => BandNormalized(sed, bright) }
+    override val create: ValidatedInput[BandNormalized[T]] = {
+      // we can't update lucuma-core thus we need to default to somthing
+      val modelSed = sed.toOption.map(_.toUnnormalizedSed)
+        .getOrElse(UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.O5V).validNec)
+      (modelSed, brightnesses.notMissingAndThen("brightnesses")(brightnessCreator[T])).mapN { (sed, bright) =>
+        BandNormalized(sed, bright)
+      }
+    }
```
swalker2m, line 313: In the `edit` below, we should probably switch to `sed.validateNullable` and then I guess treat an attempt to assign `null` as `05V`?

Maybe

```
s <- sed.validateNullable(_.toUnnormalizedSed).liftState
_ <- BandNormalized.set[T] := s.map(_.getOrElse(UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.O5V)))
```

modules/core/src/main/scala/lucuma/odb/api/schema/SourceProfileSchema.scala
```diff
@@ -583,7 +583,7 @@ object SourceProfileSchema {
   )(implicit ev: InputType[BandBrightnessInput[T]]): InputObjectType[BandNormalizedInput[T]] =
     InputObjectType[BandNormalizedInput[T]](
       s"BandNormalized${groupName.capitalize}Input",
-      s"""Create or edit a band normalized value with $groupName magnitude units.  Specify both "sed" and "brightnesses" when creating a new BandNormalized${groupName.capitalize}.""",
+      s"""Create or edit a band normalized value with $groupName magnitude units.  Specify "brightnesses" and an optional "sed" parameter when creating a new BandNormalized${groupName.capitalize}.""",
       List(
         InputObjectUnnormalizedSed.createRequiredEditOptional("sed", s"BandNormalized${groupName.capitalize}"),
```
swalker2m, line 588: This would switch to `nullableField("sed")`.

swalker2m (commented):
No need to change anything, but in case it comes up.

modules/core/src/main/scala/lucuma/odb/api/model/targetModel/SourceProfileModel.scala
cquiroz: Right, at the moment explore won't try to send a null but maybe we should support it for api clients?

swalker2m: Oh, well especially in that case it doesn't seem like it's worth the trouble considering tmp-api is EOL.

